PR #8, opened 2024-06-16
Miscellaneous clarifications and cleaning up

ibarakaiev:
### Contributor checklist

This is a minor PR aimed at slightly improving the readability of the code. In addition to the actual changes, I would like to propose the following (happy to implement myself, but wanted to discuss before jumping to writing code, esp since it might lead to deprecating previous names):

1. `Igniter.Code.Common.env_at_cursor/1` doesn't yet have a cursor (it gets inserted as the first step). Perhaps a more descriptive name would be `(get_)current_environment/1`?
2. The name for `Igniter.Code.Common.add_code/3` doesn't immediately indicate where the code is added. How's `insert_code/3` to explicitly indicate that it gets inserted exactly where the zipper is focused?
3. The name `Igniter.Code.Common.highest_adjacent_block/1` was a bit confusing to me (perhaps because the code nests left<->right while AST nests up<->down, and it's unclear which mode of thinking was necessary). How's something like `outer_containing_block/1`? (Separately, wouldn't it always return `defmodule`'s block? Or maybe I'm misunderstanding the functionality.)
4. `Igniter.Code.Common.maybe_move_to_singleton_block/1` was confusing to me because "singleton" also is a pattern from (trigger warning) OOP, so my first hunch was thinking about a "unique" block or something. After reading the code it seems that it simply means "block with one child". With this in mind, how's `maybe_move_to_single_child_block/1` or `maybe_move_to_single_expression_block/1`? I also still don't fully understand its purpose.

Separately, I often get confused by `move` since it's four-directional and depends on the orientation as described above. What do you think about using orientation-agnostic terms such as "ascend" (in AST-coordinates, it means go up; in code, it means to the left, indentation-wise), "descend", "next" as opposed to just "next" or "move", where possible? So, for example, `move_to_singleton_block` would become `descend_to_singleton_block`. UPD: reading it again and wondering if descend is still ambiguous and could be interpreted as moving up the AST 😐.

And finally, another thing that's a bit confusing is minor terminology like patterns vs predicates, nodes, etc. One example is `node_matches_pattern?/2` which actually accepts a zipper (which itself has a node, among other things). Another example is: in `Igniter.Code.Common.move_to/2`, it's a bit unclear what "predicate" is supposed to be.

Just wanted to flag these ideas since Igniter is still in its early stages and its API is still easy to modify/rename. Some of these suggestions probably only warrant documentation modifications, though.

### DIFF
--- a/lib/code/common.ex
+++ b/lib/code/common.ex
@@ -28,6 +28,17 @@ defmodule Igniter.Code.Common do
 
   @doc """
   Returns `true` if the current node matches the given pattern.
+
+  ## Examples:
+
+  ```elixir
+  list_zipper =
+    "[1, 2, 3]"
+    |> Sourceror.parse_string!()
+    |> Sourceror.Zipper.zip()
+
+  Common.node_matches_pattern?(list_zipper, value when is_list(value)) # true
+  ```
   """
   defmacro node_matches_pattern?(zipper, pattern) do
     quote do
@@ -62,22 +73,22 @@ defmodule Igniter.Code.Common do
   @doc """
   Adds the provided code to the zipper.
 
-  Use the `placement` to determine if the code goes `:after` or `:before` the current node.
+  Use `placement` to determine if the code goes `:after` or `:before` the current node.
 
   ## Example:
 
   ```elixir
-  existing_code = \"\"\"
+  existing_zipper = \"\"\"
   IO.inspect("Hello, world!")
   \"\"\"
   |> Sourceror.parse_string!()
+  |> Sourceror.Zipper.zip()
 
   new_code = \"\"\"
   IO.inspect("Goodbye, world!")
   \"\"\"
 
-  existing_code
-  |> Sourceror.Zipper.zip()
+  existing_zipper
   |> Igniter.Common.add_code(new_code)
   |> Sourceror.Zipper.root()
   |> Sourceror.to_string()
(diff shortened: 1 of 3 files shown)


zachdaniel:
Sorry it took me so long to get back to you, there is a lot of great stuff here! Lets get this PR merged and then move these into separate issues to discuss each one :)

zachdaniel:

zachdaniel:
1. https://github.com/ash-project/igniter/issues/11
2. I think we can leave `add_code` as it is for now. I don't think that `insert_code` is enough of an improvement to warrant changing it. (and everything in igniter/zipper does things relative to your current location)
3. https://github.com/ash-project/igniter/issues/12
4. I like `maybe_move_to_single_child_block/1`, sure. I'll make that change now. The reasoning behind this is that you can sometimes have a "value" that is wrapped in a block, i.e `{:__block__, [], [1]}`. If you want to check if "the node I'm looking at is an integer literal", you have to unwrap any single-child blocks to check that.

zachdaniel:
Finally

> And finally, another thing that's a bit confusing is minor terminology like patterns vs predicates, nodes, etc. One example is node_matches_pattern?/2 which actually accepts a zipper (which itself has a node, among other things). Another example is: in Igniter.Code.Common.move_to/2, it's a bit unclear what "predicate" is supposed to be.

I think we'll have to solve most of these with documentation. That function does take a `zipper`, but it checks if the current node at that zipper matches a pattern. So `node_matches_pattern?(zipper, predicate)` makes sense IMO.

As for `move_to`, whenever we use `move_*` in a predicate, we're talking about moving "next-wards" in the tree, which is a depth-first search unless otherwise qualified. i.e `move_to_function_call_in_current_scope` clarifies that it won't break certain conditions. `Code.move_to` is going to find the "next" occurrence in the code where the predicate matches. I *think* for those it makes sense to keep those `move_*` prefixed concepts.

PR #18, opened 2024-06-26
improvement: add proper_test_support_location/1

ibarakaiev:
### Contributor checklist

This PR adds `proper_test_support_location/1`, which returns idiomatic test support location such as `MyApp.DataCase` -> `test/support/data_case.ex`. It also inlines the tests as doctests for readability.


zachdaniel:
We should make a patcher that adds `test/support` to your libc directories, and then call it whenever you add a file into test/support.

zachdaniel:

PR #25, opened 2024-06-27
fix: version string splitting

ibarakaiev:
### Contributor checklist

This PR fixes how the dependency version string is parsed in `igniter.install`. Right now, adding a `ref` with another `@` at the end raises a match error because the first split produces three parts instead of two, making lines 201 and 214 are not reachable.


zachdaniel (approved):

zachdaniel:

Issue #24, opened 2024-06-27
Add `fetch_dependencies_before_continuing/1`


ibarakaiev:
**Is your feature request related to a problem? Please describe.**
Not really a problem, but sometimes a step might depend on fetching additional dependencies. Example: `mix igniter.install ash_money` adds `:ash_money` to the list of dependencies, then runs its installers, then `ash_money` sets up the Postgres stuff, but that also depends on `:ex_money_sql`, which needs to be fetched before proceeding. It would also be better for DX if `mix deps.get` doesn't need to be separately run once an igniter completes.

**Describe the solution you'd like**
Something like `Igniter.fetch_dependencies_before_continuing/1` would be helpful. It would print out a confirmation message that dependencies are to be installed after changes are applied, run `mix deps.get`, then proceed to the remaining steps. This might also simplify the logic in `Igniter.Util.Install`.

**Describe the alternatives you've considered**
One option is to manually add `mix deps.get` as `Igniter.add_task/3` into the pipeline, but that would deviate from the user experience of the first prompt, i.e. `mix igniter.install ...` would print a human-friendly message at first, and then just "the following tasks will be run: mix deps.get".

PR #28, opened 2024-06-29
improvement: add Igniter.apply_and_fetch_dependencies/1 and Igniter.has_changes?/1

ibarakaiev:
### Contributor checklist

This PR implements https://github.com/ash-project/igniter/issues/24 and refactors `Igniter.Util.Install` to use it. It also adds `Igniter.has_changes?/1` and changes one place where it's relevant. The logic in the installer is now as follows: if there are changes when running `Igniter.apply_and_fetch_dependencies/1`, prompt the user to accept them and install dependencies. If not, do nothing and move on. 

Open question: do we still need the `options` variable in the installer, considering the changes? (I marked it with `# TODO` in the code for now.) Marking as draft for now.

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -370,21 +370,41 @@ defmodule Igniter do
   end
 
   @doc """
-  Applies the current changes in the Igniter and fetches dependencies.
+  Applies the current changes to the `mix.exs` in the Igniter and fetches dependencies.
 
-  Returns an empty Igniter if successful.
+  Returns the remaining changes in the Igniter if successful.
+
+  ## Options
+
+  * `:error_on_abort?` - If `true`, raises an error if the user aborts the operation. Returns the original igniter if not.
   """
-  def apply_and_fetch_dependencies(igniter) do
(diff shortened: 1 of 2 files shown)


zachdaniel:
To make this function generally applicable, it has to *only* save the changes to the `mix.exs` file before doing the rest of its work. 

Additionally `has_changes?` is insufficient in its current implementation, we have to check if any of the sources is are from a `:string` or if they've been updated.

```elixir
  def has_changes?(igniter) do
     igniter.rewrite
     |> Rewrite.sources()
     |> Enum.any?(fn source -> 
       Rewrite.Source.from?(source, :string) || Rewrite.Source.updated?(source)
     end)
   end
```

However, I don't think we need this helper quite yet because the install and continue function needs to do something more like this:

```elixir
defp mix_exs_changed?(igniter) do
  case Rewrite.source(igniter.rewrite, "mix.exs") do
    {:ok, source} -> Rewrite.Source.from?(source, :string) || Rewrite.Source.changed?(source)
    _ -> false
  end
end
```

PR #33, opened 2024-07-04
fix: preserve original ordering in Util.Install

ibarakaiev:
### Contributor checklist

Depending on the order in which Mix loads tasks, the original installer ordering may be lost. For example, `mix igniter.install ash_postgres@path:../ash_postgres,ash_money@path:../ash_money` installs AshMoney before AshPostgres. This PR adds a sorting step to return the original ordering passed as arguments to `igniter.install`.

### DIFF
--- a/lib/igniter/util/install.ex
+++ b/lib/igniter/util/install.ex
@@ -55,6 +55,10 @@ defmodule Igniter.Util.Install do
       end)
       |> Stream.filter(&implements_behaviour?(&1, Igniter.Mix.Task))
       |> Enum.filter(&(Mix.Task.task_name(&1) in desired_tasks))
+      |> Enum.sort_by(
+        &Enum.find_index(desired_tasks, fn e -> e == Mix.Task.task_name(&1) end),
+        &<=/2
+      )
 
     title =
       case desired_tasks do


lib/igniter/util/install.ex
```diff
@@ -55,6 +55,10 @@ defmodule Igniter.Util.Install do
       end)
       |> Stream.filter(&implements_behaviour?(&1, Igniter.Mix.Task))
       |> Enum.filter(&(Mix.Task.task_name(&1) in desired_tasks))
+      |> Enum.sort_by(
+        &Enum.find_index(desired_tasks, fn e -> e == Mix.Task.task_name(&1) end),
+        &<=/2
+      )
```
zachdaniel, line 61: Maybe a cleaner approach here would be something like:

```elixir
   available_igniter_tasks =
      Mix.Task.load_all()
      |> Stream.map(fn item ->
        Code.ensure_compiled!(item)
        item
       end)
       |> Stream.filter(&implements_behaviour?(&1, Igniter.Mix.Task))
       |> Enum.filter(&(Mix.Task.task_name(&1) in desired_tasks))

igniter_tasks = Enum.filter(desired_tasks, &(&1 in available_igniter_tasks))
```

? Maybe not though, they kind of both aren't 100% clear what's going on. I'll leave it up to you just LMK

zachdaniel (approved):

lib/igniter/util/install.ex
ibarakaiev: This is tricky because `desired_tasks` would need to be converted from strings to modules before checking for `in available_igniter_tasks`. And the conversion is not deterministic, which is why we have `load_all() |> map() |> filter()`, right?

zachdaniel:

PR #34, opened 2024-07-04
improvement: append dependency to mix instead of prepending

ibarakaiev:
### Contributor checklist

This is a minor improvement to append dependencies to `mix.exs` instead of prepending. This change is not critical but is helpful for ordering purposes, in light of https://github.com/ash-project/igniter/pull/33.

### DIFF
--- a/lib/igniter/project/deps.ex
+++ b/lib/igniter/project/deps.ex
@@ -10,6 +10,7 @@ defmodule Igniter.Project.Deps do
   # Options
 
   - `:yes?` - Automatically answer yes to any prompts.
+  - `:append?` - Append to the dependency list instead of prepending.
   - `:error?` - Returns an error instead of a notice on failure.
   """
   def add_dep(igniter, dep, opts \\ []) do
@@ -146,7 +147,11 @@ defmodule Igniter.Project.Deps do
             end
           end
 
-        Igniter.Code.List.prepend_to_list(zipper, quoted)
+        if Keyword.get(opts, :append?, false) do
+          Igniter.Code.List.append_to_list(zipper, quoted)
+        else
+          Igniter.Code.List.prepend_to_list(zipper, quoted)
+        end
       else
         _ ->
           if opts[:error?] do


zachdaniel (commented):
I don't think I'm a fan of appending instead of prepending. In most real apps users have a slew of dependencies, and the bottom of the list is typically test/dev only dependencies.

ibarakaiev:
Fair enough, I can see how in an average case it could be more cumbersome. What about supporting an optional flag? I'm working on a little wrapper that runs installer tasks consecutively according to the order in `mix.exs`, and it'd be helpful there for consistency.

zachdaniel:
Something I've been thinking of is the idea of "dependency groups". Something I do in all of my apps looks like this:

```elixir
defp deps do
  [
    # Ash
    {:ash, ...},
    {:ash_phoenix, ...}

    # Web
    {:phoenix, ...}

    # Dev/Test
    {:credo, ...}
  ]
end
```

So what if we did a thing where, in the `Info` for an installer, it can specify a group, I.e `group: "Dev/Test"`. Then we can also put in a big list of known deps and groups into igniter, and have some heuristics like "if it starts with `ash_` or `phoenix_`" we guess those groups.

Then, if we have a group, we *append* to the group, and if not then we *prepend* to the list.

Thoughts?

zachdaniel:
Granted, the example I cited where I group them up kind of makes me wonder why your wrapper would want to care about or be dependent on the order of deps in mix.exs TBH. That seems a bit like a smell to me?

ibarakaiev:
Well, the project I'm working on looks at another Elixir app that defines a certain config, installs dependencies that are listed there, and copies the source code into the current directory (like a DX-friendly almost-zero config Igniter chain, with some additional goodies). In the example app I'm using to test that, there's both `ash_postgres` and `ash_money`, and `ash_money` has an optional add-on for Postgres, which checks whether there's `use AshPostgres.Repo` in the Repo module. I guess a true way to fix this dependency conflict would be to define dependencies for installers (i.e. `ash_money.install` may optionally depend on `ash_postgres.install`), then topologically sort the dependency graph, and run in that order (as opposed to running in the order provided in the CLI). Then it is not up to user to mess up the order, and my wrapper won't be affected by the ordering.

But I think I'd probably still prefer a flag to be able to append to the list instead, as I personally add dependencies in a certain order (similar to how you group dependency groups, but also with a certain readability order, i.e. `ash` always comes before `ash_*`).

zachdaniel:
Gotcha. In that case an option that defaults to prepending makes sense to me.

ibarakaiev:
Added the optional flag 👍

zachdaniel (approved):

PR #37, opened 2024-07-07
fix: remove Elixir from module_name_prefix

ibarakaiev:
### Contributor checklist

I do not know why, but the original implementation in some cases prepends `Elixir` to the return value in `Igniter.Code.Module.module_name_prefix/0`, whereas in some cases it doesn't. However, the new implementation empirically never prepends it.

### DIFF
--- a/lib/igniter/code/module.ex
+++ b/lib/igniter/code/module.ex
@@ -386,10 +386,7 @@ defmodule Igniter.Code.Module do
   @doc "The module name prefix based on the mix project's module name"
   @spec module_name_prefix() :: module()
   def module_name_prefix do
-    Mix.Project.get!()
-    |> Module.split()
-    |> :lists.droplast()
-    |> Module.concat()
+    Mix.Project.get!() |> Module.split() |> List.first() |> String.to_atom()
   end
 
   @doc "Moves the zipper to a defmodule call"


ibarakaiev:
This example is showcased in the generated `DataCase` here: https://github.com/ash-project/ash_postgres/pull/344. With the modification in this PR, `Elixir.` no longer appears in module name. Again, not sure why :)

lib/igniter/code/module.ex
```diff
@@ -386,10 +386,7 @@ defmodule Igniter.Code.Module do
   @doc "The module name prefix based on the mix project's module name"
   @spec module_name_prefix() :: module()
   def module_name_prefix do
-    Mix.Project.get!()
-    |> Module.split()
-    |> :lists.droplast()
-    |> Module.concat()
+    Mix.Project.get!() |> Module.split() |> List.first() |> String.to_atom()
```
zachdaniel, line 389: This isn't quite right, because if the mix project is something like `Foo.Bar.MixProject` then the prefix should be `Foo.Bar`.

zachdaniel (changes requested):
I'm not sure why `Elixir.` is being added as a prefix, but the logic of taking only the first part of the module name will produce incorrect results in certain mix projects as described.

zachdaniel:
I see what's happening here :) When you string interpolate a module, i.e `"#{module}"` it always adds the `Elixir` prefix. For this reason, we do `"#{inspect(module)}"`.

PR #38, opened 2024-07-07
improvement: support :kind in find_and_update_or_create_module/5

ibarakaiev:
### Contributor checklist

Sometimes it is necessary to create a new module of a certain kind, i.e. a test or a test support, for which we have helpers to get their idiomatic location. This PR slightly refactors `find_and_update_or_create_module/5` in a backwards-compatible way to support an optional `:kind` flag, which defaults to `:lib`, but also supports `:test` and `:test_support`.

### DIFF
--- a/lib/igniter/code/module.ex
+++ b/lib/igniter/code/module.ex
@@ -4,8 +4,29 @@ defmodule Igniter.Code.Module do
   alias Igniter.Code.Common
   alias Sourceror.Zipper
 
-  @doc "Finds a module and updates its contents wherever it is. If it does not exist, it is created with the provided contents."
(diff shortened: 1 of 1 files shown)


zachdaniel:
I like the idea of the `path` option, but I'm not a big fan of the `:kind` option, TBH. But with the `path` option you could very easily do something like `path: Module.proper_test_location(...)`. Could we start w/ just that `path` option?

zachdaniel (changes requested):
I think just the `path` name should be provided.

zachdaniel:
Going to merge this and remove the `kind` option in favor of just the `path` option :)

zachdaniel:

PR #45, opened 2024-07-11
refactor: simplify zipper usage

zachallaun:
I love just browsing and reading code, and while doing so with `Igniter`, noticed some zipper usage that couple be simplified.

Commits are atomic and can be reviewed independently. Happy to omit anything you don't want; just let me know. 🙂

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
 erlang 26.0.2
-elixir 1.17.1
+elixir 1.17.1-otp-26
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -209,6 +209,10 @@ defmodule Igniter.Code.Common do
     end
   end
 
+  @doc """
+  Replaces full module names in `new_code` with any aliases for that
+  module found in the `current_code` environment.
+  """
   def use_aliases(new_code, current_code) do
     case current_env(current_code) do
       {:ok, env} ->
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -61,8 +61,7 @@ defmodule Igniter.Code.Common do
       ast =
         unquote(zipper)
         |> Igniter.Code.Common.maybe_move_to_single_child_block()
-        |> Zipper.subtree()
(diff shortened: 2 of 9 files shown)


.tool-versions
```diff
@@ -1,2 +1,2 @@
 erlang 26.0.2
-elixir 1.17.1
+elixir 1.17.1-otp-26
```
zachallaun, line 2: Context: I had 1.17.1-otp-27 already installed, which was selected by my tool manager (mise), and then I was getting confusing errors when running any mix task because Elixir needs to be compiled with an OTP less than or equal to the one you're running.

lib/igniter/code/common.ex
```diff
@@ -205,36 +195,24 @@ defmodule Igniter.Code.Common do
...
-        upwards
-        |> Zipper.node()
-        |> case do
-          {:__block__, _, _} ->
-            case highest_adjacent_block(upwards) do
-              nil -> upwards
-              zipper -> zipper
-            end
+      %Zipper{node: {:__block__, _, _}} = upwards ->
+        highest_adjacent_block(upwards) || upwards
```
zachallaun, line 205: Alternatively:

```suggestion
        case highest_adjacent_block(upwards) do
          nil -> upwards
          highest -> highest
        end
```

zachdaniel:
Love it. Definitely some weird habits I got into when learning Zipper that made its way into a lot of this code 😆 Awesome simplifications 🙇 

Will merge once the build passes.

zachallaun:
Looking into the Dialyzer error.

zachallaun:
For posterity, the Dialyzer error was that [this clause](https://github.com/ash-project/igniter/pull/45/commits/0def6d07d3ef2a8afb0a5ab54b9c8819c2b6b1d8#diff-814c41fa971ce8a1f0c3d6ae9050c1185ec6d21133687dd9e277a684137e8e9fL182-L187) would never match because, if it got to that point, `upwards.node` would already be known to be a `{:__block__, _, _}` tuple.

zachdaniel:

PR #49, opened 2024-07-16
improvement: support tuple dependencies in igniter.install

ibarakaiev:
### Contributor checklist

This PR refactors `Igniter.Util.Install.install/3` into three different function clauses where `deps` can be either a string (separated with a comma), a list of strings, or a list of source dependencies (i.e. `{:ash, "~> 3.0"}`). This makes it possible to run `igniter.install` from a non-CLI context. This PR also adds a fourth `opts` function argument to `Igniter.Util.Install.install/3` (now `install/4`) and propagates it to `Igniter.Util.Info`. Note: I also removed `Application.ensure_all_started(:req)` since `Req` is not used anymore it seems.

### DIFF
--- a/lib/igniter/util/install.ex
+++ b/lib/igniter/util/install.ex
@@ -1,5 +1,7 @@
 defmodule Igniter.Util.Install do
-  @moduledoc false
+  @moduledoc """
+  Tools for installing packages and running their associated Igniter installers, if present.
+  """
 
   @doc """
   Installs the provided list of dependencies. `deps` can be either:
--- a/lib/igniter/util/install.ex
+++ b/lib/igniter/util/install.ex
@@ -1,6 +1,12 @@
 defmodule Igniter.Util.Install do
   @moduledoc """
-  Tools for installing packages and running their associated Igniter installers, if present.
+  Tools for installing packages and running their associated
+  installers, if present.
+
+  [!NOTE]
+  The functions in this module are not composable, and are primarily meant to
+  be used internally and to support building custom tooling on top of Igniter,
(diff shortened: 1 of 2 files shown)


zachdaniel:
Currently `Igniter.Util.Install` is a private module (`@moduledoc false` at the top). I'm open to making this public, but would want to know more about how you are using it.

ibarakaiev:
My personal use case is to install dependencies via Igniter according to a reference `mix.exs` definition in Fireside: https://github.com/ibarakaiev/fireside/blob/1f8e6f839a0fe4bcb8eee408f63feff891acf505/lib/fireside/util/install.ex#L23. I currently have a subset of the code from Igniter's source itself, but it's not very sustainable and I would prefer to let Igniter handle it. One way to approach is it to convert the dependency tuple into the format Igniter expects, but it's a bit of a Procrustean bed—so it'd be great if Igniter exposed its installation function publicly as well.

zachdaniel:
I can see that. Lets make the module public by adding some module docs in that case 👍

ibarakaiev:
✅

lib/igniter/util/install.ex
```diff
@@ -1,5 +1,7 @@
 defmodule Igniter.Util.Install do
-  @moduledoc false
+  @moduledoc """
+  Tools for installing packages and running their associated Igniter installers, if present.
```
zachdaniel, line 3: Lets add a line here about how this is not composable like other installers, and this is meant to be used if you are building custom tooling on top of igniter aside from "standard" generators/installers/patchers. Feel free to link to your project as an example.

zachdaniel (approved):

PR #50, opened 2024-07-17
improvement: only show executed installers

ibarakaiev:
### Contributor checklist

This PR changes the title of the result of running all installers to only include the installers that actually ran. It also only runs the installers if there is at least one task that has an installer.


zachdaniel (approved):

zachdaniel:

PR #53, opened 2024-07-18
improvement: add Function.argument_equals?/3

ibarakaiev:
### Contributor checklist

This PR adds `Igniter.Code.Function.argument_equals?/3` to make it easy to check argument equality. It also adds a note to `Function.argument_matches_pattern?/3` to clarify that equality checks should be done with `argument_equals?/3`.

### DIFF
--- a/lib/igniter/code/function.ex
+++ b/lib/igniter/code/function.ex
@@ -384,7 +384,7 @@ defmodule Igniter.Code.Function do
         &Igniter.Code.Common.nodes_equal?(&1, term)
       )
     else
-      raise "The provided zipper is not a function call."
+      false
     end
   end
 
--- a/lib/igniter/code/function.ex
+++ b/lib/igniter/code/function.ex
@@ -7,7 +7,11 @@ defmodule Igniter.Code.Function do
   alias Igniter.Code.Common
   alias Sourceror.Zipper
 
-  @doc "Returns `true` if the argument at the provided index exists and matches the provided pattern"
+  @doc """
+  Returns `true` if the argument at the provided index exists and matches the provided pattern
+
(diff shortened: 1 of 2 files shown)


lib/igniter/code/function.ex
```diff
@@ -367,6 +371,14 @@ defmodule Igniter.Code.Function do
     end
   end
 
+  def argument_equals?(zipper, index, term) do
```
zachdaniel, line 374: can we add a `@doc` and `@spec` here?

ibarakaiev:
Added a `@doc` and a `@spec` as well as a check to make sure the zipper is actually a function call.

lib/igniter/code/function.ex
```diff
@@ -367,6 +371,23 @@ defmodule Igniter.Code.Function do
...
+  @spec argument_equals?(Zipper.t(), integer(), any()) :: boolean()
+  def argument_equals?(zipper, index, term) do
+    if function_call?(zipper) do
+      Igniter.Code.Function.argument_matches_predicate?(
+        zipper,
+        index,
+        &Igniter.Code.Common.nodes_equal?(&1, term)
+      )
+    else
+      raise "The provided zipper is not a function call."
```
zachdaniel, line 387: Let's just return `false` here, and leave it up to them to validate that they are in a function call.
```suggestion
      false
```

ibarakaiev: Sure, that works. Just didn't want to return an `:error` (as it was, probably mistakenly, returned in `configures?/3` before) as it introduces a subtle bug since it's truthy.

zachdaniel:

Issue #51, opened 2024-07-18
`Igniter.Project.Config.configures?/3` doesn't work in all general cases


ibarakaiev:
Facing some issues with retrieving configuration in Igniter. Specifically, `Igniter.Project.Config.configures?/3` seems to return `:error` where it shouldn't. I can make a PR for it to return `false` but I think there are a few more issues, outlined below.

There are two versions of `Config.config/2,3`, one that accepts just the `root_key` and `opts` and one that also accepts a given `key` for the `root_key`: https://hexdocs.pm/elixir/main/Config.html#config/2. Tbh I'm not entirely sure how they're different since one can achieve the same functionality with both:
```elixir
config :my_app, [{Test, [hello: :world]}]
config :my_app, Test, hello: :world
```
I guess it's a convenience and a convention to only define configuration up to two nested levels deep. This means that `Igniter.Project.Config.configures/3` needs to search for either `Config.config/2` or `Config.config/3`. In principle, it is easy to assume that one needs to search for `Config.config/2` if the provided `config_path` is a single element, and `Config.config/3` when it has more than one element. But in reality `Config.config/2` may be nested too (like in the example above), meaning that when there's more than one element, `Config.configures/3` needs to search for both versions of the `config`. Right now what happens is `Config.configures/3` searches for `Config.config/3` when `config_path` is one element, and for `Config.config/2` when `config_path` is more than one element. This seems incorrect, and at least in my case it doesn't quite work.

One additional complication that needs to be accounted for is that a certain key may be configured twice, like in the code examples in the official docs linked above. Therefore, a certain nested key may not be configured in the first `Config.config/2,3` match but may be configured in a later one. So there needs to be some kind of a `move_right()` + recursion if there was no nested match in a given `Config.config/2,3` match.

Finally, I found the function signature of `Config.configures?/3` a bit tricky to reason about since the `app_name` is last in the list, whereas in `config/2,3` it is first and is called `root_key`.

Considering the above, I wanted to propose the following change:

1. Add `Config.configures_root_key?(zipper, root_key)` which will check if either `config :my_app, _` or `config :my_app, _, _` is present.
1. Add `Config.configures_key?(zipper, root_key, key)` which will check if either `config :my_app, key: _` or `config :my_app, :key, _` is present.
2. Add wrappers to pass a file instead of a zipper instead.
[...]

PR #54, opened 2024-07-18
fix: configures?/3 -> configures_key & configures_root_key

ibarakaiev:
### Contributor checklist

This PR splits (and deprecates) `Igniter.Project.Config.configures?/3` into `Igniter.Project.Config.configures_key/4` and `Igniter.Project.Config.configures_root_key/3` to allow passing just the root key, atom keys, and list keys (as keyword paths).

Additionally, it addresses some of the issues discussed in #51, specifically, it:
- moves `Igniter.Code.Keyword.keyword_has_path?/2` inside the matching function in `Igniter.Code.Function.move_to_function_call_in_current_scope/4` to make sure deeply nested keys defined in a _later_ (that is, not first) run of the same root key configuration is checked. Previously, it was matching on the root key and checking it within a `case`, so there was a possibility it would match on the first occurrence, fail to find a nested key, and return `false` immediately.
- makes sure only boolean values are ever returned.
- supports both `config :root_key, :key, 1` as well as `config :root_key, key: 1` notations (as well as their deeply nested counterparts).
- makes the order and naming more consistent with Elixir's `Config.config/2,3`, that is, using `root_key` (instead of `app_name`) and using it earlier in the arguments.

### DIFF
--- a/lib/igniter/code/function.ex
+++ b/lib/igniter/code/function.ex
@@ -453,7 +453,7 @@ defmodule Igniter.Code.Function do
         end
       end
     else
-      :error
+      raise "The passed Zipper is not a function."
     end
   end
 
--- a/lib/igniter/code/function.ex
+++ b/lib/igniter/code/function.ex
@@ -453,7 +453,7 @@ defmodule Igniter.Code.Function do
         end
       end
     else
-      raise "The passed Zipper is not a function."
+      false
     end
   end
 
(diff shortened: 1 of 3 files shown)


zachdaniel:
Thanks for the PR. Will review tomorrow 🙇

ibarakaiev:
Thanks! I ended up deviating from the original proposal a bit in light of your comments, so it's essentially a renaming of `configures?/3` (to fix the order) and a new helper function for empty paths.

zachdaniel (approved):

zachdaniel:

PR #57, opened 2024-07-23
WIP: read from the mix file

### LINKED ISSUE #5: Check source for application module in `Igniter.Application.add_new_child/2`
At the moment, we check the existing application, but this could be out of date due to some existing patch.

Here is where we check it:

```elixir
    to_perform =
      case project.application()[:mod] do
        nil -> {:create_an_app, Igniter.Code.Module.module_name("Application")}
        {mod, _} -> {:modify, mod}
        mod -> {:modify, mod}
      end
```

This should pull from the source.

sigu:
### Contributor checklist

closes #5

### DIFF
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -16,10 +16,8 @@ defmodule Igniter.Project.Application do
   @doc "Adds a new child to the `children` list in the application file"
   @spec add_new_child(Igniter.t(), module() | {module, term()}) :: Igniter.t()
   def add_new_child(igniter, to_supervise) do
-    project = Mix.Project.get!()
-
     to_perform =
-      case project.application()[:mod] do
+      case mod_definition(igniter) do
         nil -> {:create_an_app, Igniter.Code.Module.module_name("Application")}
         {mod, _} -> {:modify, mod}
         mod -> {:modify, mod}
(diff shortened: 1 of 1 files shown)


zachdaniel:
To add a test, we'd need to add new tests. Some of the other tests show similar examples, like the dependency tests IIRC.

lib/igniter/project/application.ex
```diff
@@ -16,10 +16,8 @@ defmodule Igniter.Project.Application do
   @doc "Adds a new child to the `children` list in the application file"
   @spec add_new_child(Igniter.t(), module() | {module, term()}) :: Igniter.t()
   def add_new_child(igniter, to_supervise) do
-    project = Mix.Project.get!()
-
     to_perform =
-      case project.application()[:mod] do
+      case mod_definition(igniter) do
         nil -> {:create_an_app, Igniter.Code.Module.module_name("Application")}
```
zachdaniel, line 21: We need to have an `:error` case here, where we couldn't parse it. Then we'd want to prompt the user.
Something like:

```
Could not parse project configuration. Do we need to create an application? y/n
# if no
What is the name of the application?
```

zachdaniel:
Closing as this has been fixed in `main`.

PR #63, opened 2024-08-03
Replacing the Application module with the generic application module that gets derived from the function args

sinisijacob:
Please let me know if this change is not correct. I'm just starting to make use of this library and I think I noticed an issue with the do_add_child function. It seems that the module that should be getting changed should be the generic application that the user passes in, which is not necessarily the Application module.

I am very new to Igniter and Elixir in general, just trying to help if there's an actual issue here.

### Contributor checklist

- [  ] Bug fixes include regression tests
- [  ] Features include unit/acceptance tests

### DIFF
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -64,7 +64,7 @@ defmodule Igniter.Project.Application do
       end
 
     Igniter.update_elixir_file(igniter, path, fn zipper ->
-      with {:ok, zipper} <- Igniter.Code.Module.move_to_module_using(zipper, Application),
+      with {:ok, zipper} <- Igniter.Code.Module.move_to_module_using(zipper, application),
            {:ok, zipper} <- Igniter.Code.Function.move_to_def(zipper, :start, 2),
            {:ok, zipper} <-
              Igniter.Code.Function.move_to_function_call_in_current_scope(


lib/igniter/project/application.ex
```diff
@@ -64,7 +64,7 @@ defmodule Igniter.Project.Application do
       end
 
     Igniter.update_elixir_file(igniter, path, fn zipper ->
-      with {:ok, zipper} <- Igniter.Code.Module.move_to_module_using(zipper, Application),
+      with {:ok, zipper} <- Igniter.Code.Module.move_to_module_using(zipper, application),
```
zachdaniel, line 67: What this should be doing is moving to a module that contains this

```elixir
use Application
```

With this change, it would be looking for something like:

```elixir
use MyApp.Application
```

Which would be incorrect

zachdaniel:
Thank you for the PR! This change is not correct unfortunately so I will close this PR. However, I'm curious what brought on the PR. Did you encounter a bug using igniter?

sinisijacob:
Thank you for the clarification, I'm still on the steep part of the learning curve.
I have not encountered a bug in Igniter, I am still trying to get my head wrapped around how it works. I was reading through the code to find some examples to work off of and saw what looked like it might be an issue. I really admire the work you're doing with this project and am planning on making use of it for a project I am working on.

PR #64, opened 2024-08-03
check for extendable_block?/1 in maybe_move_to_single_child_block/1

ibarakaiev:
### Contributor checklist

Creating a PR following the conversation in Igniter discord.

### Context

If I have something like `config :my_app, SomeModule, key1: [version: 1]` and I use `Igniter.Project.Config.configure(...)` to update the value to another integer (say, 2), under the hood it changes the version to:
```elixir
{:__block__, [trailing_comments: [], leading_comments: [], token: "1", ...], [2]}
```
but running `Sourceror.to_string/1` on the above produces `config my_app,  SomeModule, key1: [version: 1]` again because of the `token: "1"`. (The behavior is similar in Elixir's `Macro` itself.)

The current approach is implementing @zachdaniel's suggestion but so far it breaks several tests (and most likely some things elsewhere too).

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -384,15 +384,19 @@ defmodule Igniter.Code.Common do
 
   def maybe_move_to_single_child_block(zipper) do
     case zipper.node do
-      {:__block__, _, [_]} ->
-        zipper
-        |> Zipper.down()
-        |> case do
-          nil ->
-            zipper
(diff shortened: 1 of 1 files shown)


zachdaniel:
Added to your change here: https://github.com/ash-project/igniter/pull/69

zachdaniel:

PR #65, opened 2024-08-03
improvement: support "notices"

ibarakaiev:
### Contributor checklist

This PR adds the support for "notices", which is the programmable output that will be printed once an Igniter runs successfully. They are printed in green, and are intended for things like "next steps" and such.

UPD: this PR also pretties up the output by surrounding relevant warning and notice sections with blank lines where necessary. The old "Notices" has also been renamed to "Warnings".

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -3,7 +3,7 @@ defmodule Igniter do
   Tools for generating and patching code into an Elixir project.
   """
 
-  defstruct [:rewrite, issues: [], tasks: [], warnings: [], assigns: %{}, moves: %{}]
+  defstruct [:rewrite, issues: [], tasks: [], warnings: [], notices: [], assigns: %{}, moves: %{}]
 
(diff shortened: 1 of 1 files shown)


zachdaniel:
There are places where you edit a file that you can return `{:warning, ...}` and `{:error, ...}`, we should make those support `{:notice, ...}` as well. Otherwise, LGTM!

ibarakaiev:
What should the return format be? `{:notice, {zipper, notice}}` or something like that? For warnings and errors, a zipper is not modified, but for notices it is. I almost feel like notices should only ever be added for things like "next steps" or "please read the README here: ..." so I'm not sure what would be a use case for returning a notice tuple like that 🤔

zachdaniel:
🤔 yeah, okay good point. Lets leave it :) So this is ready to go?

ibarakaiev:
Yeah, I think so!

zachdaniel:

PR #77, opened 2024-08-13
improvement: add failing config tests

ibarakaiev:
### Contributor checklist

Adding two failing tests for the `Config.configure/6` function. The first has to do with nesting (seems like there's still a bug lurking around). The second one has to do with injecting quoted code.

Currently, the first test fails with:
```
error: syntax error before: '=>'
         │
       2 │ config :level1, :level2, level3: ["hello1" => "world1"]
```

and the second fails with no matching function clause error.


zachdaniel:
Will merge and see what can be done.

PR #89, opened 2024-09-22
fix: Accept an empty string when prompting

jimsynz:
The prompt `[Yn]` would indicate that `Y` is the default action, therefore we should accept an empty string as an indicator that the user wants us to continue.

### DIFF
--- a/lib/igniter/util/io.ex
+++ b/lib/igniter/util/io.ex
@@ -4,7 +4,7 @@ defmodule Igniter.Util.IO do
   @doc "Prompts the user for yes or no, repeating the prompt until a satisfactory answer is given"
   def yes?(prompt) do
     case String.trim(Mix.shell().prompt(prompt <> " [Yn]")) do
-      yes when yes in ["y", "Y", "yes", "YES"] ->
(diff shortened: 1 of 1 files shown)


lib/igniter/util/io.ex
```diff
@@ -4,7 +4,7 @@ defmodule Igniter.Util.IO do
   @doc "Prompts the user for yes or no, repeating the prompt until a satisfactory answer is given"
   def yes?(prompt) do
     case String.trim(Mix.shell().prompt(prompt <> " [Yn]")) do
-      yes when yes in ["y", "Y", "yes", "YES"] ->
+      yes when yes in ["y", "Y", "yes", "YES", ""] ->
```
zachdaniel, line 7: we actually just changed this explicitly for that not be the case. I hadn't considered that the capital Y implies that it's the default, but it intentionally requires explicit acceptance.

zachdaniel (changes requested):

lib/igniter/util/io.ex
jimsynz: okay then we need to fix the case of the prompt.

PR #107, opened 2024-10-21
Fix gen_migration :overwrite

leandrocp:
Fix a typo and _try_ to add a test although I'm not sure that test is correctly handling `igniter.assigns[:test_files]`.

### DIFF
--- a/lib/igniter/libs/ecto.ex
+++ b/lib/igniter/libs/ecto.ex
@@ -71,7 +71,7 @@ defmodule Igniter.Libs.Ecto do
                 gen_migration(igniter, repo, name <> "_1", opts)
             end
 
-          :ovewrwrite ->
+          :overwrite ->
             Igniter.create_new_file(
               igniter,
               file,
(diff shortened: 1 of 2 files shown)


zachdaniel:
What do you meant that it's not correctly handling `:test_files`? I'll maybe see more when the tests run.

leandrocp:
I had a `apply_igniter!/1` after the first `gen_migration` because I assume that was necessary to actually create the file and then assert that `:overwrite` works properly but seems like that removes the file from `:test_files` so I'm not sure the test is correct, but I guess it's fine.

zachdaniel:
Hmm...yeah thats not right. I'll look into it

zachdaniel:
It looks like it's in test_files to me 🤔

zachdaniel:
Looks like its working as intended to me

leandrocp:
Well then never mind. I was doing some tests and might have changed something unexpected. Still learning how the internal parts work :)

PR #109, opened 2024-10-25
Add `mix_project/1` and `phoenix_project/1`

leandrocp:
- Deprecate `test_project/1` in favor of `mix_project/1` (just because it made sense to make it more clear)
- Format test files (required because Phoenix files are not formatted)

Missing in this PR:
- Umbrella
- Other opts such as --module, --no-ecto, etc

I've tested these changes on https://github.com/BeaconCMS/beacon/blob/main/test/mix/tasks/install_test.exs and https://github.com/BeaconCMS/beacon/blob/main/test/mix/tasks/gen_site_test.exs with success, all tests passed without changes.


leandrocp:
I'm not sure why tests are failing, I can't reproduce locally 🤔

lib/igniter/test.ex
```diff
@@ -344,128 +393,226 @@ defmodule Igniter.Test do
...
+  # TODO: opts[:umbrella]
+  # TODO: maybe support other options such as --module, --no-ecto, etc
+  defp add_phoenix_new(opts) do
+    # FIXME: base path
+    app_name = to_string(opts[:app_name] || :test)
+    source_base_path = Path.expand("../../deps/phx_new/templates", __DIR__)
+
+    project =
+      app_name
+      |> Phx.New.Project.new([])
```
zachdaniel, line 570: We can't use this like this if we have only a dev/test dependency on `phx_new`. It would cause compilation errors/warnings in any application that wants to use it. So to do this we'd need to make `phx_new` an optional *non* dev/test dependency of igniter, and then do something like this:

```elixir
if Code.ensure_loaded?(Phx.New) do
  def phoenix_project(opts \\ []) do
    ...current implementation
  end
else
  def phoenix_project(opts \\ []) do
     raise "You must include the `phx_new` dependency to generate to use `#{inspect(__MODULE__)}.phoenix_project/1`"
  end
end
```

zachdaniel:
Hey @leandrocp whats the status here? Just got back from vacation last week, so still catching up :)

leandrocp:
Hey @zachdaniel I'm closing this PR but I'm wondering what you think about deprecating `test_project` in favor of `mix_project` to make it more explicit.

zachdaniel:
Oops, I forgot about this PR 😅 sorry. Been all over the place. Yes, feel free to PR that deprecation.

zachdaniel:
Add an upgrader with a function renamed if possible as well.

leandrocp:
> Oops, I forgot about this PR 😅 sorry. Been all over the place. Yes, feel free to PR that deprecation.

No worries, this one was mixed with that Phoenix installer 😅

PR #111, opened 2024-10-26
improvement: resolve project app names set using a module attribute

zachallaun:
This PR also fixes two bugs in `Common.expand_literal/1`. Previously, the following would be returned:

```elixir
iex(1)> ":foo" |> Sourceror.parse_string!() |> Sourceror.Zipper.zip() |> Igniter.Code.Common.expand_literal()
{:ok, {:__block__, [trailing_comments: [], leading_comments: [], line: 1, column: 1], [:foo]}}
iex(2)> "@foo" |> Sourceror.parse_string!() |> Sourceror.Zipper.zip() |> Igniter.Code.Common.expand_literal()
{:ok,
 {:@, [trailing_comments: [], leading_comments: [], line: 1, column: 1],
  [{:foo, [trailing_comments: [], leading_comments: [], line: 1, column: 2], nil}]}}
```

With this PR:

```elixir
iex(1)> ":foo" |> Sourceror.parse_string!() |> Sourceror.Zipper.zip() |> Igniter.Code.Common.expand_literal()
{:ok, :foo}
iex(2)> "@foo" |> Sourceror.parse_string!() |> Sourceror.Zipper.zip() |> Igniter.Code.Common.expand_literal()
:error
```

### Contributor checklist

### DIFF
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -29,7 +29,7 @@ defmodule Igniter.Project.Application do
          zipper <- Igniter.Code.Common.rightmost(zipper),
          true <- Igniter.Code.List.list?(zipper),
          {:ok, zipper} <- Igniter.Code.Keyword.get_key(zipper, :app),
-         {:ok, app_name} when is_atom(app_name) <- Igniter.Code.Common.expand_literal(zipper) do
+         {:ok, app_name} when is_atom(app_name) <- expand_app_name(zipper) do
       app_name
     else
       _ ->
@@ -41,6 +41,57 @@ defmodule Igniter.Project.Application do
         """
     end
   end
+
+  defp expand_app_name(zipper) do
+    with :error <- Common.expand_literal(zipper),
+         :error <- expand_attribute(zipper) do
+      :error
+    end
+  end
+
+  defp expand_attribute(zipper) do
+    with {:@, _, [{attr, _, nil}]} <- zipper.node,
+         {:ok, zipper} <- move_up_to_defmodule_body(zipper) do
+      zipper
+      |> Stream.unfold(fn
+        %Zipper{} = zipper -> {zipper, Zipper.left(zipper)}
+        nil -> nil
+      end)
+      |> Enum.find_value(:error, fn
+        %Zipper{node: {:@, _, [{^attr, _, [_]}]}} = zipper ->
+          zipper
+          |> Zipper.down()
+          |> Zipper.down()
+          |> Common.expand_literal()
+
+        _ ->
+          nil
+      end)
+    else
+      _ ->
+        :error
+    end
+  end
+
+  defp move_up_to_defmodule_body(zipper) do
+    # when there are > 1 top-level nodes in a defmodule body, they will
+    # be 4 levels down
+    case move_upwards_n_times(zipper, 4) do
+      {:ok, maybe_defmodule} ->
+        if Igniter.Code.Function.function_call?(maybe_defmodule, :defmodule, 2) do
+          {:ok, zipper}
+        else
+          zipper |> Zipper.up() |> move_up_to_defmodule_body()
+        end
+
+      :error ->
+        :error
+    end
+  end
+
+  defp move_upwards_n_times(nil, _), do: :error
+  defp move_upwards_n_times(zipper, 0), do: {:ok, zipper}
+  defp move_upwards_n_times(zipper, n), do: zipper |> Zipper.up() |> move_upwards_n_times(n - 1)
 
   @doc "Returns the name of the application module."
   def app_module(igniter) do
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -51,47 +51,29 @@ defmodule Igniter.Project.Application do
 
   defp expand_attribute(zipper) do
     with {:@, _, [{attr, _, nil}]} <- zipper.node,
-         {:ok, zipper} <- move_up_to_defmodule_body(zipper) do
+         {:ok, zipper} <- find_prev(zipper, &match?({:@, _, [{^attr, _, [_]}]}, &1.node)) do
       zipper
-      |> Stream.unfold(fn
-        %Zipper{} = zipper -> {zipper, Zipper.left(zipper)}
-        nil -> nil
-      end)
-      |> Enum.find_value(:error, fn
-        %Zipper{node: {:@, _, [{^attr, _, [_]}]}} = zipper ->
-          zipper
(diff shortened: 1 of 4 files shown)


lib/igniter/code/module.ex
```diff
@@ -48,6 +48,13 @@ defmodule Igniter.Code.Module do
     )
   end
 
+  # TODO: Decide if this is the correct API and make public
```
zachallaun, line 51: This seemed like a potentially useful API (given a node, find the containing defmodule), but I didn't want to shoehorn it into this PR. If you'd prefer this not be here, I'm happy to make it a private function where it's used.

lib/igniter/project/application.ex
```diff
@@ -42,6 +42,56 @@ defmodule Igniter.Project.Application do
...
+          |> Zipper.down()
+          |> Zipper.down()
+          |> Common.expand_literal()
+
+        _ ->
+          nil
+      end)
+    else
+      other ->
+        {:error_other, other}
```
zachallaun, line 86: Oops, cruft from debugging, will fix and rebase.

lib/igniter/code/module.ex
zachdaniel: Seems very useful indeed! Go for it 😁

zachallaun: Just to be clear: Should I make it public and document it in this PR, or in a separate PR after this is merged?

lib/igniter/project/application.ex
```diff
@@ -42,6 +42,52 @@ defmodule Igniter.Project.Application do
...
+  defp expand_app_name(zipper) do
+    with :error <- Common.expand_literal(zipper),
+         :error <- expand_attribute(zipper) do
+      :error
+    end
+  end
+
+  defp expand_attribute(zipper) do
+    with {:@, _, [{attr, _, nil}]} = attr_getter <- zipper.node,
+         %{} = getter_range <- Sourceror.get_range(attr_getter),
```
zachdaniel, line 54: I don't think that we can rely on this unfortunately. Igniter is predicated on the fact that other things may have written to this zipper before this function was called, and so may not have line number info.  i think we have to "move out to the defmodules scope" and then move up until we find the module attribute declaration.

lib/igniter/code/module.ex
zachdaniel: 😅sorry, in this PR is good 😀

lib/igniter/project/application.ex
zachallaun: Ah, yes, you're right. If that's the case, I won't need the `Module.move_up_to_defmodule/1` helper. If it's alright with you, I'd prefer to remove it from this PR and I can submit a followup shortly after.

lib/igniter/code/module.ex
zachallaun: See [here](https://github.com/ash-project/igniter/pull/111#discussion_r1817875500); plan to do this in another PR since this function isn't necessary for this PR anymore.

lib/igniter/project/application.ex
```diff
@@ -42,6 +42,57 @@ defmodule Igniter.Project.Application do
...
+         :error <- expand_attribute(zipper) do
+      :error
+    end
+  end
+
+  defp expand_attribute(zipper) do
+    with {:@, _, [{attr, _, nil}]} <- zipper.node,
+         {:ok, zipper} <- move_up_to_defmodule_body(zipper) do
+      zipper
+      |> Stream.unfold(fn
```
zachdaniel, line 56: Could this be replaced by a `move_left` function with a predicate?

zachdaniel:
Looking great. One small question added. Also curious if you'd rather get the other one in before this.

zachdaniel:
We can ignore the CI issue, that is a misconfiguration in our CI that I need to address.

lib/igniter/project/application.ex
zachallaun: It could, but I don't think `move_left` exists yet, unless I somehow missed it.

We could add `move_left` in the other PR, get it merged in, then update this PR to use it.

zachdaniel: I'd vote for that I think :) Making more ergonomic (and idiomatic) public APIs to let people build with igniter is high priority at the moment, so I'm always looking for opportunities for that.

zachallaun:
Marked as a draft for now because this PR now relies on #113, which hasn't been merged yet, so tests are now failing. Once it's merged, I'll rebase and mark as ready for review.

lib/igniter/project/application.ex
zachallaun: Definitely a nice improvement using `move_left`! 08db3e5ef1b48096135f88caf1a14ca25d98cb79

zachallaun:
I believe this is ready to go pending a final review, @zachdaniel :)

lib/igniter/project/application.ex
```diff
@@ -42,6 +42,47 @@ defmodule Igniter.Project.Application do
...
+    else
+      _ ->
+        :error
+    end
+  end
+
+  defp move_up_to_defmodule_body(zipper) do
+    # when there are > 1 top-level nodes in a defmodule body, they will
+    # be 4 levels down
+    case move_upwards_n_times(zipper, 4) do
```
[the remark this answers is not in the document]
zachallaun: Also, I'm realizing in looking at this that I should use the new `Common.move_upwards` that was implemented in the other PR.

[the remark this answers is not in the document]
zachdaniel: 🤔 thinking about it, do we necessarily need to check that we're in the module body? We can tell the difference between the AST that assigns to a module attribute and the AST for accessing it, right? So we could theoretically just `move_left/up` (a.k.a prev, right?) until we find assignment to that @attr?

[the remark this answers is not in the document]
zachallaun: 👍 Good idea, will refactor.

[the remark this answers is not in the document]
zachdaniel: I did make a conscious choice to treat everything is "down/right", i.e `move_to`. I think this simplifies a lot of the confusing parts of using zippers, and is 99% of the time what those writing generators/patchers will need to do. But it does mean that you can't just "reverse" most operations, so we will need conscious stuff like "move_back_to". I hate that name, so something else 😆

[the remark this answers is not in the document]
zachallaun: Refactored; good call on this, I do think it's more understandable.

zachdaniel (approved):

zachdaniel:

Issue #112, opened 2024-10-26
`Igniter.Code.Common`: Support integer in `move_right` and `move_upwards` in addition to function predicate


zachallaun:
Currently, `move_right` has a corresponding `nth_right` that moves the zipper to the right a given number of times, returning `:error` if it can't move that many times. At the moment, there isn't a corresponding function for `move_upwards`.

I'd like to propose that:

- `move_right` and `move_upwards` additionally accept an integer as the second argument, in which case they'd act the same as `nth_right` and the hypothetical `nth_upwards`.
- `nth_right` be deprecated in favor of passing an integer to `move_right`.

It's my opinion that this API is more ergonomic and easier to discover.

If this proposal is declined, I'd also accept adding an `nth_upwards` as a consolation prize. 🙂

zachallaun:
Closed by #113.

PR #113, opened 2024-10-26
`Igniter.Code.Common`: Support integer argument in `move_right/2` and `move_upwards/2` and add `move_left/2`

zachallaun:
This PR implements #112, if accepted.

Note: If #111 is merged first, this PR should be rebased to replace a [private function](https://github.com/ash-project/igniter/pull/111/files#diff-7a0262dc445950a410b05838f4415a4532f854ceebf504cacd6b9c0280000ad6R92-R94) used there with a call to `Igniter.Code.Common.move_upwards(zipper, 4)`. If this PR is merged first, #111 can be rebased instead.

### Contributor checklist

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -45,22 +45,36 @@ defmodule Igniter.Code.Common do
   end
 
   @doc """
-  Moves to the next node that matches the predicate, going upwards.
+  Moves a zipper upwards.
+
+  If the second argument is a predicate function, it will be called on the zipper and then
+  move upwards until the predicate returns `true`.
+
+  If the second argument is a non-negative integer, it will move upwards that many times if
+  possible, returning `:error` otherwise.
   """
-  @spec move_upwards(Zipper.t(), (Zipper.t() -> boolean())) :: {:ok, Zipper.t()} | :error
-  def move_upwards(zipper, pred) do
+  @spec move_upwards(Zipper.t(), non_neg_integer() | (Zipper.t() -> boolean())) ::
+          {:ok, Zipper.t()} | :error
+  def move_upwards(zipper, pred_or_n)
+
+  def move_upwards(%Zipper{} = zipper, pred) when is_function(pred, 1) do
     if pred.(zipper) do
       {:ok, zipper}
     else
-      case Zipper.up(zipper) do
-        nil ->
-          :error
-
-        next ->
-          move_upwards(next, pred)
+      case move_upwards(zipper, 1) do
+        :error -> :error
+        {:ok, next} -> move_upwards(next, pred)
       end
     end
   end
+
+  def move_upwards(%Zipper{} = zipper, n) when is_integer(n) and n >= 0 do
+    do_nth_upwards(zipper, n)
+  end
+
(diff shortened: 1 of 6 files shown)


lib/igniter/code/common.ex
```diff
@@ -766,22 +777,10 @@ defmodule Igniter.Code.Common do
     end
   end
 
-  @doc "Moves the zipper right n times, returning `:error` if it can't move that many times."
+  @deprecated "Use `move_right/2` instead, passing an integer as the second argument."
   @spec nth_right(Zipper.t(), non_neg_integer()) :: {:ok, Zipper.t()} | :error
```
zachdaniel, line 781: So, we don't *have* to do this, but this is a good case to add an updater that will update this in other apps automatically. We can use `Igniter.Refactor.Rename` to do this, since the new function is a superset of the existing function.

zachdaniel: See this: https://github.com/ash-project/igniter/blob/main/lib/mix/tasks/igniter.upgrade_igniter.ex

zachallaun: Sounds good! Will add that and implement `move_left` as discussed in #111.

zachallaun:
@zachdaniel Implemented `move_left/2` and updated `igniter.upgrade_igniter`. I think this should be good to go. :)

lib/mix/tasks/igniter.upgrade_igniter.ex
```diff
@@ -42,7 +42,8 @@ defmodule Mix.Tasks.Igniter.UpgradeIgniter do
     upgrades =
       %{
         "0.3.66" => [&code_module_parse_to_project_module_parse/2],
-        "0.3.71" => [&code_module_parse_to_project_module_parse/2]
+        "0.3.71" => [&code_module_parse_to_project_module_parse/2],
+        "0.3.76" => [&code_common_nth_right_to_move_right/2]
```
zachallaun, line 46: I chose the version that I expect this to release in; is that the correct thing to do? (We're on 0.3.75 as of now.)

lib/mix/tasks/igniter.upgrade_igniter.ex
```diff
@@ -60,7 +61,19 @@ defmodule Mix.Tasks.Igniter.UpgradeIgniter do
       arity: 1
     )
     |> Igniter.add_notice(
-      "Igniter.Code.Module.parse/1 was deprecated in favor of `Igniter.Project.Module.parse/1"
```
zachallaun, line 63: This is highlighting weird, but the only change to this existing upgrade function is removing the stray backtick.

lib/mix/tasks/igniter.upgrade_igniter.ex
zachdaniel: Yep!

zachdaniel (approved):

zachdaniel:
Awesome work! It's great to have more folks who understand the internals (as you naturally do 😂) helping expand and refine these APIs 🙇‍♂️

zachdaniel:

PR #120, opened 2024-10-29
improvement: add `Igniter.Test.diff/2`

zachallaun:
I found myself writing [some tests](https://github.com/zachallaun/mneme/blob/main/test/mix/tasks/mneme.install_test.exs) where I wanted the diff as a string, but couldn't use `capture_io` and `Igniter.Test.puts_diff/2` because it included terminal escape codes for color.

This PR adds `Igniter.Test.diff/2` to support such a case.

### Contributor checklist

### DIFF
--- a/lib/igniter/test.ex
+++ b/lib/igniter/test.ex
@@ -7,8 +7,7 @@ defmodule Igniter.Test do
   ## Starting point
 
   All of the files of an empty mix project are added by default.
-  You can specify more or overwrite the default files by passing a map of
-  file paths to their contents.
+  You can specify more or overwrite files with the `:files` option.
 
   ## Limitations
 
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):

lib/igniter/test.ex
```diff
@@ -41,45 +40,50 @@ defmodule Igniter.Test do
...
 
-      diff ->
-        if opts[:label] do
-          IO.puts("#{opts[:label]}\n\n#{diff}")
-        else
-          IO.puts(diff)
-        end
+    case igniter |> diff(only: only, color?: true) |> String.trim() do
+      "" -> IO.puts("#{prefix}no changes!")
+      diff -> IO.puts("#{prefix}\n\n#{diff}")
```
zachallaun, line 60: Oops, this will actually add 2 extra newlines even if you don't specify a label. Will fix.

zachallaun: Fixed.

zachdaniel:
Hmm....that dialyzer error sucks, but it doesn't look to be in `main` so must be related to your changes?

zachallaun:
Thanks Dialyzer! ed11ad38812bbc0b21c2c9c54b4a3d1f2eeb34e1

PR #122, opened 2024-10-29
Ignitier.Project.Application.app_module/1 returning tuple

davydog187:
- [x] Figure out how to reproduce

### Contributor checklist

### DIFF
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -93,17 +93,13 @@ defmodule Igniter.Project.Application do
          true <- Igniter.Code.List.list?(zipper),
          {:ok, zipper} <- Igniter.Code.Keyword.get_key(zipper, :mod) do
       case Igniter.Code.Common.expand_literal(zipper) do
-        # {:ok, {app_module, _}} ->
-        #   {:ok, app_module}
-
-        {:ok, app_module} ->
+        {app_module, _} ->
           {:ok, app_module}
 
         :error ->
           try do
-            zipper.node
-            |> Code.eval_quoted()
-            |> elem(0)
+            {{module, _}, _} = Code.eval_quoted(zipper.node)
+            module
           rescue
             _ ->
               reraise """
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -92,9 +92,9 @@ defmodule Igniter.Project.Application do
          {:ok, zipper} <- Igniter.Code.Common.move_right(zipper, &Igniter.Code.List.list?/1),
          true <- Igniter.Code.List.list?(zipper),
          {:ok, zipper} <- Igniter.Code.Keyword.get_key(zipper, :mod) do
-      case dbg(Igniter.Code.Common.expand_literal(dbg(zipper))) do
-        {app_module, _} ->
-          {:ok, app_module}
+      case Igniter.Code.Common.expand_literal(zipper) do
+        {:ok, {app_module, _}} ->
+          app_module
 
         :error ->
           try do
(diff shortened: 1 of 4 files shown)


zachdaniel:
Alright, I led you a bit astray during your livestream :)

`Igniter.Project.Application.app_module` is for getting the application module, i.e the one that has `use Application`. It looks in

```elixir
def application() do
  [
    module: {Module, []}
  ]
end
```

So to reproduce, add that function and key in your test. Then, to fix, we want to fix two things:

1. the branch that returns `{:ok, module}` should just return `module`
2. *both* branches need to match on the `{Mod, opts}` format that could be present.

davydog187:
@zachdaniel ok i fixed it, but its hitting the "unhappy" path of doing the Code eval, so I don't think this is the correct fix.

It's definitely mergeable but you may want to take a closer look at the implementation

lib/igniter/project/application.ex
```diff
@@ -90,14 +90,13 @@ defmodule Igniter.Project.Application do
          true <- Igniter.Code.List.list?(zipper),
          {:ok, zipper} <- Igniter.Code.Keyword.get_key(zipper, :mod) do
       case Igniter.Code.Common.expand_literal(zipper) do
```
zachdaniel, line 92: This function can return `{:ok, app_module}` or `{:ok, {app_module, opts}}`, so we need to handle that.

lib/igniter/project/application.ex
```diff
@@ -90,14 +90,13 @@ defmodule Igniter.Project.Application do
...
-        {:ok, app_module} ->
+        {app_module, _} ->
           {:ok, app_module}
 
         :error ->
           try do
-            zipper.node
-            |> Code.eval_quoted()
-            |> elem(0)
+            {{module, _}, _} = Code.eval_quoted(zipper.node)
```
zachdaniel, line 98: This could return a `module` or `{app_module, opts}` as well, so need to handle that here.

zachdaniel: We also need to handle the possibility of it being `nil` (since an app isn't required), causing this function to return `nil`.

lib/igniter/project/application.ex
```diff
@@ -89,18 +89,17 @@ defmodule Igniter.Project.Application do
       |> Sourceror.Zipper.zip()
 
     with {:ok, zipper} <- Igniter.Code.Function.move_to_def(zipper, :application, 0),
-         zipper <- Igniter.Code.Common.rightmost(zipper),
+         {:ok, zipper} <- Igniter.Code.Common.move_right(zipper, &Igniter.Code.List.list?/1),
```
zachdaniel, line 92: I think we still need `rightmost` here.

zachdaniel: That would be the "thing that this function returns"

davydog187: Well either there's a bug in rightmost or thats not what happens. This was to handle the edge case where there was code before the keyword list. `rightmost/1` wasn't handling that

zachdaniel: 🤔 okay, that will have to be addressed in that case. What this should mean is that if you did this:

```elixir
def application() do
  []

  [
    application: {Module, []}
  ]
end
```

that test would fail (because the current implementation stops at the first list). So the bug you're referring to will have to be fixed some other way. I can try to find some time today to fix the bug w/ `rightmost`, if you wouldn't mind pushing it back up to use `rightmost`.

davydog187: I added a test to verify https://github.com/ash-project/igniter/pull/128

zachdaniel (approved):

zachdaniel:

PR #123, opened 2024-10-29
`Igniter.Code.Common.replace_code/2`: Don't leave zipper at parent when extending blocks

zachallaun:
While working on something else, I noticed that `replace_code/2` would leave the zipper at the parent block in many cases. Here's an example of a test that fails on main:

```elixir
  test "replaces in blocks" do
    zipper =
      """
      block do
        one()
        two()
        three()
      end
      """
      |> Sourceror.parse_string!()
      |> Zipper.zip()
      |> Zipper.search_pattern("two()")

    expected = """
    block do
      one()
      replaced()
      three()
    end\
    """

    replaced = Common.replace_code(zipper, "replaced()")
    assert expected == replaced |> Zipper.topmost_root() |> Sourceror.to_string()
    assert {:replaced, _, []} = replaced.node
  end
```

The first assertion will succeed, but the second will fail, because `replaced.node` is now the block containing the three calls.

This PR fixes that bug and additionally refactors `replace_code/2`. I'd suggest reviewing commits in order: the first commit adds passing tests, the second commit adds failing assertions and fixes the bug, and the third commit refactors the function.

### Contributor checklist

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -491,7 +491,13 @@ defmodule Igniter.Code.Common do
       {head, tail} =
         Enum.split(upwards_code, index)
 
-      Zipper.replace(upwards, {:__block__, [], head ++ to_insert ++ Enum.drop(tail, 1)})
+      {:ok, zipper} =
+        upwards
+        |> Zipper.replace({:__block__, [], head ++ to_insert ++ Enum.drop(tail, 1)})
+        |> Zipper.down()
+        |> move_right(index)
+
+      zipper
     else
       _ ->
         with nil <- Zipper.up(zipper),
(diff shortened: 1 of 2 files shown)


lib/igniter/code/common.ex
```diff
@@ -468,95 +468,48 @@ defmodule Igniter.Code.Common do
       v
   end
 
-  def replace_code(zipper, code) when is_binary(code) do
+  def replace_code(%Zipper{} = zipper, code) when is_binary(code) do
```
zachdaniel, line 471: unrelated to this, but we need to remove the places that take a binary and guess that it is code. We have places that accept `{:code, code}`, but we should probably replace it with an "unrealistic value", `{:__code__, code}` or something.

zachdaniel:

Issue #126, opened 2024-10-30
Proposal: Parse `argv` by default and store in `%Igniter{}`


zachallaun:
[Original discussion on Discord](https://discord.com/channels/711271361523351632/1253324141491261472/1301195781042208799)

Currently, when you create an `Igniter.Mix.Task`, the `argv` is passed as a second argument to the `igniter/2` callback and you are expected to parse it using `Igniter.Mix.Task.positional_args!/1` and `Igniter.Mix.Task.options!/1`. You then manage the returned arguments and options yourself.

This proposal is for changing the way Igniter handles `argv`; instead of expecting users to parse themselves, Igniter will (by default) parse into a new `%Igniter.Mix.Task.Args{}` struct accessible through `igniter.args`. The struct fields will include: `:positional`, `:options`, `:argv` (what you get after positional args are removed), and `:raw` (what you get from Mix). To support custom argument parsing, an optional `parse_argv/2` callback will be added.

zachallaun:
Closed by #131.

PR #131, opened 2024-10-31
Parse `argv` by default and store in `Igniter` struct (#126)

zachallaun:
Work in progress implementation of #126.

- [x] Add `%Igniter.Mix.Task.Args{}` struct and add an `:args` field to `%Igniter{}`
- [x] Introduce an optional `parse_argv/2` callback to the `Igniter.Mix.Task` behaviour
- [x] Update the task runner to parse args, call `igniter/1` if available, and fall back to `igniter/2`
- [x] Update task composition to parse args and set the `:args` field prior to running a composed task
- [x] Warn if both `igniter/1` and `igniter/2` are defined
- [x] Raise if neither `igniter/1` nor `igniter/2` are defined
- [x] Bug: `compile_task/4` needs to reset the igniter's args after running the composed task
- [x] Update tests
- [x] Update Igniter tasks
- [x] Update generators
- [x] Update docs and guides
- [x] Update `mix igniter.upgrade_igniter` to automatically move `igniter/2` to `igniter/1` for simple cases

### Contributor checklist

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -3,7 +3,16 @@ defmodule Igniter do
   Tools for generating and patching code into an Elixir project.
   """
 
-  defstruct [:rewrite, issues: [], tasks: [], warnings: [], notices: [], assigns: %{}, moves: %{}]
+  defstruct [
+    :rewrite,
+    issues: [],
+    tasks: [],
+    warnings: [],
+    notices: [],
+    assigns: %{},
+    moves: %{},
+    args: %Igniter.Mix.Task.Args{}
+  ]
 
   alias Sourceror.Zipper
 
@@ -14,7 +23,8 @@ defmodule Igniter do
           warnings: [String.t()],
           notices: [String.t()],
           assigns: map(),
-          moves: %{optional(String.t()) => String.t()}
+          moves: %{optional(String.t()) => String.t()},
+          args: Igniter.Mix.Task.Args.t()
         }
 
   @type zipper_updater :: (Zipper.t() -> {:ok, Zipper.t()} | {:error, String.t() | [String.t()]})
@@ -333,16 +343,17 @@ defmodule Igniter do
   Finds the `Igniter.Mix.Task` task by name and composes it (calls its `igniter/2`) into the current igniter.
   If the task doesn't exist, a fallback implementation may be provided as the last argument.
   """
-  def compose_task(igniter, task, argv \\ [], fallback \\ nil)
+  def compose_task(igniter, task, argv \\ nil, fallback \\ nil)
 
   def compose_task(igniter, task, argv, fallback) when is_atom(task) do
+    argv = argv || igniter.args.argv
     Code.ensure_compiled!(task)
 
-    if function_exported?(task, :igniter, 2) do
+    if function_exported?(task, :igniter, 1) or function_exported?(task, :igniter, 2) do
       if !task.supports_umbrella?() && Mix.Project.umbrella?() do
         add_issue(igniter, "Cannot run #{inspect(task)} in an umbrella project.")
       else
-        task.igniter(igniter, argv)
+        Igniter.Mix.Task.configure(igniter, task, argv)
       end
     else
       if is_function(fallback) do
@@ -361,6 +372,8 @@ defmodule Igniter do
   end
 
   def compose_task(igniter, task_name, argv, fallback) do
+    argv = argv || igniter.args.argv
+
     if igniter.issues == [] do
       task_name
       |> Mix.Task.get()
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -342,14 +342,41 @@ defmodule Igniter do
   @doc """
   Finds the `Igniter.Mix.Task` task by name and composes it with `igniter`.
 
-  This will call the task's `igniter/1` (or `igniter/2`) callback, setting
-  `igniter.args` using the current `igniter.args.argv_flags`. This prevents
-  composed tasks from accidentally consuming positional arguments. If you
-  wish the composed task to access positional arguments, you must call this
-  function with `igniter.args.argv` as the `argv` argument.
-
   If the task doesn't exist, a `fallback` function may be provided. This
   function should accept and return the `igniter`.
+
+  ## Argument handling
+
+  This function calls the task's `igniter/1` (or `igniter/2`) callback, setting
+  `igniter.args` using the current `igniter.args.argv_flags`. This prevents
+  composed tasks from accidentally consuming positional arguments. If you
+  wish the composed task to access additional arguments, you must explicitly
+  pass an `argv` list.
+
+  Additionally, you must declare other tasks you are composing with in your
+  task's `Igniter.Mix.Task.Info` struct using the `:composes` key. Without
+  this, you'll see unexpected argument errors if a flag that a composed task
+  uses is passed without you explicitly declaring it in your `:schema`.
+
+  ## Example
+
+      def info(_argv, _parent) do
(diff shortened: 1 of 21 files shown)


lib/igniter.ex
```diff
@@ -333,16 +343,17 @@ defmodule Igniter do
   Finds the `Igniter.Mix.Task` task by name and composes it (calls its `igniter/2`) into the current igniter.
   If the task doesn't exist, a fallback implementation may be provided as the last argument.
   """
-  def compose_task(igniter, task, argv \\ [], fallback \\ nil)
+  def compose_task(igniter, task, argv \\ nil, fallback \\ nil)
```
zachdaniel, line 346: I think we might want to require that they provide the argv here still? What I worry about is accidentally making it easy to misuse the positional arguments to your task, that should very likely not be passed into a composed task.

lib/mix/task.ex
```diff
@@ -89,10 +106,37 @@ defmodule Igniter.Mix.Task do
...
+        options = options!(argv_flags)
+        %Args{positional: positional, options: options, argv: argv, argv_flags: argv_flags}
+      end
+
       defoverridable supports_umbrella?: 0, info: 2, installer?: 0
     end
   end
 
+  @doc false
+  def configure(igniter, task_module, argv) do
```
zachdaniel, line 121: maybe `configure_and_run`?

lib/igniter.ex
zachallaun: So what you're saying is that it's more likely that the correct thing here is to use `igniter.args.argv_flags`, but we should leave it up to the call site? So the pattern we should advertise in the docs will be:

```elixir
igniter
|> Igniter.compose_task("some.other.task", igniter.args.argv_flags)
```

zachdaniel: Yeah, although...now that you put it that way, maybe we should just make the default `igniter.args.argv_flags` 😆 then the default behavior is also the most likely correct behavior.

zachallaun: 👍 That sounds good to me. I'll update the documentation for `compose_task` to include that.

zachallaun: Updated this in [0810f8b](https://github.com/ash-project/igniter/pull/131/commits/0810f8b649055f6df8d75e3b8534ced4fc60818c), please take a look at those changes and let me know your thoughts!

lib/igniter.ex
```diff
@@ -330,32 +340,56 @@ defmodule Igniter do
...
+  Finds the `Igniter.Mix.Task` task by name and composes it with `igniter`.
+
+  This will call the task's `igniter/1` (or `igniter/2`) callback, setting
+  `igniter.args` using the current `igniter.args.argv_flags`. This prevents
+  composed tasks from accidentally consuming positional arguments. If you
+  wish the composed task to access positional arguments, you must call this
+  function with `igniter.args.argv` as the `argv` argument.
+
+  If the task doesn't exist, a `fallback` function may be provided. This
+  function should accept and return the `igniter`.
```
zachdaniel, line 352: LGTM! It is a common pattern to decide add your own positional arguments to a composed task, so perhaps we should show that as an example:

```elixir
Igniter.compose_task("task", ["positional", "argument"] ++ igniter.args.argv_glags)
```

We use this in the Ash installer when the `--example` flag is given, we compose the `ash.gen.resource` command twice.

lib/igniter.ex
```diff
@@ -330,32 +340,83 @@ defmodule Igniter do
...
+        }
+
+      def igniter(igniter) do
+        igniter
+        # other.task1 will see igniter.args.argv_flags as its args
+        |> Igniter.compose_task("other.task1")
+        # other.task2 will see an additional arg and flag
+        |> Igniter.compose_task("other.task2", ["arg", "--flag"] ++ igniter.argv.argv_flags)
+      end
+
```
zachallaun, line 379: Added this section in response to your other comment. This is based on my experimental understanding of how it works, but please check the above for correctness :)

zachallaun:
@zachdaniel I think the only missing thing is to update `mix igniter.upgrade_igniter`. Otherwise, I think this is in a good state to review.

zachallaun:
Updated `mix igniter.upgrade_igniter` to handle `def igniter(igniter, _argv)` and the default, existing generated tasks.

Would be good to test this locally on some existing projects.

lib/igniter/code/common.ex
```diff
@@ -470,13 +470,18 @@ defmodule Igniter.Code.Common do
...
-        |> Enum.reduce(zipper, &Zipper.insert_right(&2, &1))
-        |> replace_node(new_code_head)
+        new_code
+        # We insert to the left and explicitly replace the node with new_code_last
+        # so that the last node retains the metadata of the node we're replacing.
+        # That metadata includes things like the number of trailing newlines.
+        |> Enum.reduce(zipper, &Zipper.insert_left(&2, &1))
+        |> replace_node(new_code_last)
+
+      {:ok, zipper} = move_left(zipper, length(new_code))
```
zachallaun, line 484: This is a fix for this test, which fails on main:

```elixir
test "keeps newlines" do
  zipper =
    """
    block do
      one()

      two()
    end
    """
    |> Sourceror.parse_string!()
    |> Zipper.zip()
    |> Zipper.search_pattern("one()")

  expected =
    """
    block do
      replaced1()
      replaced2()

      two()
    end\
    """

  replaced = zipper |> Common.replace_code("replaced1()\nreplaced2()\n")
  assert {:replaced1, _, []} = replaced.node
  assert expected == replaced |> Zipper.topmost_root() |> Sourceror.to_string()
end
```

On main, the block looks like this after `replace_code`:

```elixir
block do
  replaced1()

  replaced2()
  two()
end
```

I encountered this while updating the `upgrade_igniter` task.

lib/igniter/upgrades/igniter.ex
```diff
@@ -0,0 +1,78 @@
...
+      {:ok, zipper}
+    else
+      _ -> :error
+    end
+  end
+
+  defp generated_argv_usage?(zipper) do
+    with ^zipper <- Zipper.search_pattern(zipper, "{arguments, argv} = positional_args!(argv)"),
+         {:ok, zipper} <- Common.move_right(zipper, 1),
+         ^zipper <- Zipper.search_pattern(zipper, "options = options!(argv)") do
```
zachdaniel, line 63: I think this could have some false positives, like if there was code in between these two lines, right? Because `search_pattern` uses `next`. So I think we probably want something like a "matches pattern" check, to see if this particular node matches the pattern.

zachdaniel:
🔥 This is epic 🔥 I had one comment on the upgrader. After that is addressed, I'll try out the ugprader against all of my packages, and if it looks good we can get it merged.

lib/igniter/upgrades/igniter.ex
zachallaun: I'll write a test to confirm, but since it's matching `^zipper`, it only matches if the current zipper node is returned, so I think we should be good?

zachdaniel: Something like `true <- Zipper.matches_pattern?(zipper, "...")` would be a nice addition to `Zipper` I think.

zachallaun: > Something like `true <- Zipper.matches_pattern?(zipper, "...")` would be a nice addition to `Zipper` I think.

Yep, agreed! I was looking for something like that originally and figured this out as a sort of stop-gap.

zachdaniel:
I'm going to merge this to main given that it's got green tests for everything. I'll test the upgrader against main on other projects.

zachdaniel:

PR #144, opened 2024-11-05
fix: pass dot_formatter when updating rewrite sources

### LINKED ISSUE #143: Installer not respecting `.formatter.exs`
**Describe the bug**
Installer is not respecting the applied `.formatter.exs` config.

The Beacon installer does add the `:beacon` dep into `.formatter.exs` which [exports](https://github.com/BeaconCMS/beacon/blob/main/.formatter.exs) 2 functions as `:locals_without_parens` so the resulting list of deps is `[:beacon, :ecto, :ecto_sql, :phoenix]` as you can see in the diff:

```
2   - |  import_deps: [:ecto, :ecto_sql, :phoenix],
  2 + |  import_deps: [:beacon, :ecto, :ecto_sql, :phoenix],
```

But the applied changes are not using those deps:

```
   19 + |  scope "/" do
   20 + |    pipe_through([:browser])
   21 + |    beacon_site("/", site: :my_site)
   22 + |  end
```

_Neither pipe_through nor beacon_site should have parens._

**To Reproduce**
```sh
mix archive.install hex phx_new
mix archive.install hex igniter_new
mix igniter.new my_app --install beacon@github:BeaconCMS/beacon --with phx.new --beacon.site my_site
```

**Expected behavior**
Load `:import_deps` and respect its rules.

** Runtime
- erlang 27
- elixir 1.17.2
- igniter 0.4.5
- igniter_new 0.4.11

zachallaun:
Closes #143

I'd like to wait for @leandrocp to confirm that this branch works for him before merging 👍 

### Contributor checklist

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -268,8 +268,9 @@ defmodule Igniter do
           try do
             Rewrite.update!(
               igniter.rewrite,
-              Rewrite.Source.update(
+              update_source(
                 source,
+                igniter,
                 :quoted,
                 Sourceror.Zipper.topmost_root(zipper),
                 by: :configure
@@ -563,7 +564,7 @@ defmodule Igniter do
           _ ->
             ""
             |> Rewrite.Source.Ex.from_string(path: path)
-            |> Rewrite.Source.update(:content, contents, by: :file_creator)
+            |> update_source(igniter, :content, contents, by: :file_creator)
         end
 
       %{igniter | rewrite: Rewrite.put!(igniter.rewrite, source)}
@@ -586,7 +587,7 @@ defmodule Igniter do
             {true,
              ""
              |> Rewrite.Source.Ex.from_string(path)
-             |> Rewrite.Source.update(:file_creator, :content, contents)}
+             |> update_source(igniter, :content, contents, by: :file_creator)}
         end
 
       %{igniter | rewrite: Rewrite.put!(igniter.rewrite, source)}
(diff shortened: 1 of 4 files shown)


lib/igniter.ex
```diff
@@ -586,7 +587,7 @@ defmodule Igniter do
             {true,
              ""
              |> Rewrite.Source.Ex.from_string(path)
-             |> Rewrite.Source.update(:file_creator, :content, contents)}
+             |> update_source(igniter, :content, contents, by: :file_creator)}
```
zachallaun, line 590: I believe the previous call format was already incorrect and should have been `by: :file_creator` -- something that slipped by in the update to Rewrite 1.0?

The next change is the same.

test/igniter/project/formatter_test.exs
```diff
@@ -0,0 +1,49 @@
...
+
+            test "1" do
+              expect Foo.add(x, y), do: x + y
+            end
+          end
+          """
+        }
+      )
+      |> Igniter.Project.Deps.add_dep({:mimic, "~> 1.7"})
+      |> Igniter.Project.Formatter.import_dep(:mimic)
```
zachallaun, line 22: This is a little hacky, but I needed a dep we already had that exports some locals_without_parens, and Mimic was the one I found.

leandrocp:
@zachallaun confirmed it's working as expected!

PR #148, opened 2024-11-08
refactor: handle io formatting more uniformly

zachallaun:
### Contributor checklist


lib/igniter.ex
```diff
@@ -1507,68 +1495,82 @@ defmodule Igniter do
...
-          end)
-
-        Mix.shell().info("\n" <> notices)
-    end
+  @doc false
+  def display_warnings(igniter, title) do
+    igniter.warnings
+    |> Enum.reverse()
+    |> Enum.map(fn error ->
+      ["* ", :yellow, format_error(error)]
```
barnabasJ, line 1513: Do you prefer using the atoms directly over using the `IO.ANSI.<color>` functions?

zachallaun:
> a general thought I had when going through this was if issues should be printed to stderr instead of stdio

@barnabasJ Good thought, but it's unfortunately a little tricky such that I don't want to include that change in this PR. Maybe worth opening an issue?

The two problems are:

1. We use `Mix.shell()` to print, but we can't use `Mix.shell().error(...)` because we do some custom formatting, and it would override everything, making it all red.
2. There isn't a guarantee that `Mix.shell()` is `Mix.Shell.IO`, but we could check whether it is, and in that case just do an `IO.puts(:stderr, ...)` for error cases. If `Mix.shell()` *isn't* a `Mix.Shell.IO`, we could just use the `error` function of whatever it is.

That's just enough change that I don't want to shoehorn it into this PR, but I think it's a reasonable think to address (or consider addressing) down the line.

barnabasJ:
> > a general thought I had when going through this was if issues should be printed to stderr instead of stdio
> 
> @barnabasJ Good thought, but it's unfortunately a little tricky such that I don't want to include that change in this PR. Maybe worth opening an issue?
> 
> The two problems are:
> 
>     1. We use `Mix.shell()` to print, but we can't use `Mix.shell().error(...)` because we do some custom formatting, and it would override everything, making it all red.
> 
>     2. There isn't a guarantee that `Mix.shell()` is `Mix.Shell.IO`, but we could check whether it is, and in that case just do an `IO.puts(:stderr, ...)` for error cases. If `Mix.shell()` _isn't_ a `Mix.Shell.IO`, we could just use the `error` function of whatever it is.
> 
> 
> That's just enough change that I don't want to shoehorn it into this PR, but I think it's a reasonable think to address (or consider addressing) down the line.

I also thought this could be done later if we wanted, so I approved the PR. 

I'll open an issue to track it.

PR #149, opened 2024-11-12
feat: add swoosh module

barnabasJ:
needed for https://github.com/team-alembic/ash_authentication/pull/835

### DIFF
--- a/lib/igniter/libs/swoosh.ex
+++ b/lib/igniter/libs/swoosh.ex
@@ -10,9 +10,9 @@ defmodule Igniter.Libs.Swoosh do
   end
 
   @doc "Moves to the use statement in a module that matches `use Swoosh.Mailer`"
-  @spec move_to_mailer_use(Igniter.t(), Sourceror.Zipper.t()) ::
+  @spec move_to_mailer_use(Sourceror.Zipper.t()) ::
           :error | {:ok, Sourceror.Zipper.t()}
-  def move_to_mailer_use(_igniter, zipper) do
+  def move_to_mailer_use(zipper) do
     Igniter.Code.Function.move_to_function_call(zipper, :use, 2, fn zipper ->
       Igniter.Code.Function.argument_equals?(
         zipper,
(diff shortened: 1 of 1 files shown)


lib/igniter/libs/swoosh.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Igniter.Libs.Swoosh do
+  @moduledoc "Codemods & utilities for working with Swoosh"
+
+  @doc "Lists all mailers found in the project."
```
zachallaun, line 4:
```suggestion
  @doc "Lists all project modules that call `use Swoosh.Mailer`."
```

IMO, best to be very explicit here.

lib/igniter/libs/swoosh.ex
```diff
@@ -0,0 +1,24 @@
...
+  def list_mailers(igniter) do
+    Igniter.Project.Module.find_all_matching_modules(igniter, fn _mod, zipper ->
+      move_to_mailer_use(igniter, zipper) != :error
+    end)
+  end
+
+  @doc "Moves to the use statement in a module that matches `use Swoosh.Mailer`"
+  @spec move_to_mailer_use(Igniter.t(), Sourceror.Zipper.t()) ::
+          :error | {:ok, Sourceror.Zipper.t()}
+  def move_to_mailer_use(_igniter, zipper) do
```
zachallaun, line 15: I have mixed feelings on requiring an unused argument here.

It's consistent in that all the functions in `Libs.Phoenix` and `Libs.Ecto` require it, *but* all those functions actually use the argument. Maybe this function starts using it in the future in order to be more intelligent or efficient, but I somewhat doubt it. 

We should probably either:

1. Remove the argument and make it `move_to_mailer_use/1`, or
2. Change the signature to `move_to_mailer_use(%Igniter{}, zipper)` to at least ensure the right thing is being passed in, even if it's not being used.

I think 1. is preferable (we can always deprecate it later), but I could be convinced of 2.

barnabasJ: I'm fine with 1. I just copied over some of the code from the phoenix module last week and made it work. 
I should have taken another look at it today before pushing.

zachallaun (approved):

PR #151, opened 2024-11-12
feat: propagate expand_env? option to Igniter.Code.Common.add_code/3

cfbender:
### Contributor checklist

We ran into an issue when using igniter with [styler](https://hexdocs.pm/styler/0.8.0/readme.html), where the formatting would move the `@behaviour` tag above the aliases, and then would be incorrectly aliased. I found that using the `expand_env?` argument fixes this, and propagated the option up to `add_code/3`. 

I made an opinionated change here to convert it to an `opts` list, which creates a breaking change for its usage, happy to change if this is too breaking!

Also happy to close this in order to discuss if this doesn't fit the project goals.

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -193,10 +193,14 @@ defmodule Igniter.Code.Common do
     end
   end
 
+  @type add_code_opts :: [placement: :after | :before, expand_env?: boolean()]
   @doc """
   Adds the provided code to the zipper.
 
-  Use `placement` to determine if the code goes `:after` or `:before` the current node.
+  ## Options
+
+  - `:placement` - `:after` | `:before`. Determines if the code goes `:after` or `:before` the current node. Defaults to `:after`.
+  - `:expand_env?` - boolean. Whether or not to read the current file to use it's aliases for added code. Defaults to `false`.
 
   ## Example:
 
@@ -226,20 +230,29 @@ defmodule Igniter.Code.Common do
   \"\"\"
   ```
   """
-  @spec add_code(Zipper.t(), String.t() | Macro.t(), :after | :before) :: Zipper.t()
-  def add_code(zipper, new_code, placement \\ :after)
+  @spec add_code(Zipper.t(), String.t() | Macro.t(), add_code_opts() | atom()) :: Zipper.t()
+  def add_code(zipper, new_code, opts \\ [placement: :after, expand_env?: true])
 
-  def add_code(zipper, new_code, placement) when is_binary(new_code) do
+  def add_code(_zipper, _new_code, placement) when is_atom(placement) do
+    raise ArgumentError, """
+    The `placement` argument is deprecated. Use an opts list of `placement: #{inspect(placement)}` instead.
+    """
+  end
+
+  def add_code(zipper, new_code, opts) when is_binary(new_code) do
     code = Sourceror.parse_string!(new_code)
 
-    add_code(zipper, code, placement)
+    add_code(zipper, code, opts)
   end
 
-  def add_code(zipper, new_code, placement) do
-    do_add_code(zipper, new_code, placement)
+  def add_code(zipper, new_code, opts) do
+    do_add_code(zipper, new_code, opts)
   end
 
-  defp do_add_code(zipper, new_code, placement, expand_env? \\ true) do
+  defp do_add_code(zipper, new_code, opts) do
+    expand_env? = Keyword.get(opts, :expand_env?, true)
+    placement = Keyword.get(opts, :placement, :after)
+
     new_code =
       if expand_env? do
         use_aliases(new_code, zipper)
@@ -976,7 +989,7 @@ defmodule Igniter.Code.Common do
     Process.put(:elixir_code_diagnostics, {[], false})
 
     zipper
-    |> do_add_code({:__cursor__, [], []}, :after, false)
+    |> do_add_code({:__cursor__, [], []}, placement: :after, expand_env?: false)
     |> Zipper.topmost_root()
     |> Sourceror.to_string()
     |> String.split("__cursor__()", parts: 2)
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -195,7 +195,6 @@ defmodule Igniter.Code.Common do
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/igniter/code/common.ex
```diff
@@ -226,20 +230,29 @@ defmodule Igniter.Code.Common do
...
   """
-  @spec add_code(Zipper.t(), String.t() | Macro.t(), :after | :before) :: Zipper.t()
-  def add_code(zipper, new_code, placement \\ :after)
+  @spec add_code(Zipper.t(), String.t() | Macro.t(), add_code_opts() | atom()) :: Zipper.t()
+  def add_code(zipper, new_code, opts \\ [placement: :after, expand_env?: true])
+
+  def add_code(_zipper, _new_code, placement) when is_atom(placement) do
+    raise ArgumentError, """
+    The `placement` argument is deprecated. Use an opts list of `placement: #{inspect(placement)}` instead.
+    """
```
zachallaun, line 239: Let's deprecate this instead of raising, so this should change to something like:

```elixir
def add_code(zipper, new_code, placement) when is_atom(placement) do
  Logger.warning("""
  Passing an atom as the third argument to `Igniter.Code.Common.add_code/3` is deprecated in favor of an options list.
  
  Instead of:
  
      Igniter.Code.Common.add_code(zipper, new_code, #{inspect(placement)})
  
  You should now write:
  
      Igniter.Code.Common.add_code(zipper, new_code, placement: #{inspect(placement)})
  """)

  add_code(zipper, new_code, placement: placement)
end
```

lib/igniter/code/common.ex
```diff
@@ -226,20 +230,29 @@ defmodule Igniter.Code.Common do
   \"\"\"
   ```
   """
-  @spec add_code(Zipper.t(), String.t() | Macro.t(), :after | :before) :: Zipper.t()
-  def add_code(zipper, new_code, placement \\ :after)
+  @spec add_code(Zipper.t(), String.t() | Macro.t(), add_code_opts() | atom()) :: Zipper.t()
+  def add_code(zipper, new_code, opts \\ [placement: :after, expand_env?: true])
```
zachallaun, line 234: We need to specify the defaults when they're extracted anyways (which you're already doing), so this might as well be an empty list to avoid the defaults ever getting out-of-sync.

```suggestion
  def add_code(zipper, new_code, opts \\ [])
```

zachallaun (changes requested):

zachallaun:
Not required, but if we really wanted to go the extra mile, we could update the `igniter.upgrade_igniter` task to automatically correct all existing usage of `add_code(..., ..., :before | :after)` to `add_code(..., ..., placement: :before | :after)` 🙂

lib/igniter/code/common.ex
```diff
@@ -226,20 +230,29 @@ defmodule Igniter.Code.Common do
   \"\"\"
   ```
   """
-  @spec add_code(Zipper.t(), String.t() | Macro.t(), :after | :before) :: Zipper.t()
-  def add_code(zipper, new_code, placement \\ :after)
+  @spec add_code(Zipper.t(), String.t() | Macro.t(), add_code_opts() | atom()) :: Zipper.t()
```
zachallaun, line 233: A few thoughts on this spec:

- While we'll *technically* handle the atom-3rd-argument case, it'll be deprecated, so I'm fine with Dialyzer failing and only including the list of opts in the spec.
- IIRC, the best practice is to specify kw option types as `[opt()]` as opposed to `opts()`, so we could change this to `[add_code_opt()]`, but...
- Since the options aren't shared, IMO it's nicer to co-locate them, which makes them visible in hover docs in your editor, for example:

```suggestion
  @spec add_code(Zipper.t(), String.t() | Macro.t(), [opt]) :: Zipper.t()
        when opt: {:placement, :after | :before} | {:expand_env?, boolean()}
```

cfbender:
> Not required, but if we really wanted to go the extra mile, we could update the `igniter.upgrade_igniter` task to automatically correct all existing usage of `add_code(..., ..., :before | :after)` to `add_code(..., ..., placement: :before | :after)` 🙂

I started looking at this and it looks like it may be a bit complex for me since our usage of igniter has been pretty basic so far. Happy to keep working on it - just not sure how easy it would be for someone more experienced to jam on it

lib/igniter/code/common.ex
cfbender: I like this much more thank you! Fixed in [`eec48a9` (#151)](https://github.com/ash-project/igniter/pull/151/commits/eec48a91e45d90c49a67e02621c1bf23d5cf5f32)

lib/igniter/code/common.ex
cfbender: Good call! fixed in [`71f1574` (#151)](https://github.com/ash-project/igniter/pull/151/commits/71f1574077e7374056352ff1528a1d8ed449cbd3)

lib/igniter/code/common.ex
cfbender: Ah yeah I wasn't quite sure on best practice here - fixed in [`9f79055` (#151)](https://github.com/ash-project/igniter/pull/151/commits/9f79055b35b474f81670652481cb791ee97768e3)!

zachallaun:
> I started looking at this and it looks like it may be a bit complex for me since our usage of igniter has been pretty basic so far. Happy to keep working on it - just not sure how easy it would be for someone more experienced to jam on it

No worries! I may take a crack at it later and will mention you in the PR in case you're interested in how it turns out.

test/igniter/code/common_test.exs
```diff
@@ -128,6 +129,81 @@ defmodule Igniter.Code.CommonTest do
...
+                   [Foo.Bar.Baz.Blart]
+                 end
+                 """)
+      else
+        _ ->
+          flunk("Should not reach this case")
+      end
+    end
+
+    test "raises on deprecated placement argument" do
```
zachallaun, line 169:
```suggestion
    test "warns on deprecated placement argument" do
```

cfbender: not sure what y'alls typical workflow is but I edited that commit in a rebase - good catch!

zachallaun: Rebase is fine! For a nice, atomic change like this, I prefer to squash everything anyways when I merge, but I do appreciate a nice, clean commit history regardless :)

PR #166, opened 2024-12-06
Fix parse_argv callback should be overridable

maxnordlund:
Otherwise an overridden `parse_argv/1` has no effect, as pointed out by dialyzer.

### Contributor checklist

Not sure how to test this TBH.

### DIFF
--- a/lib/mix/task.ex
+++ b/lib/mix/task.ex
@@ -123,7 +123,7 @@ defmodule Igniter.Mix.Task do
         %Args{positional: positional, options: options, argv: argv, argv_flags: argv_flags}
       end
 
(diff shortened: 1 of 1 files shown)


zachdaniel:
I'm actually not sure the original intention was to have this as an overridable callback. The reason that it is a callback at all is because the `options` and `positional_args` functions are defined by the task. @zachallaun thoughts on this change?

zachallaun:
Only able to look at this briefly, but IIRC it was intended (and is documented) as overridable/optional. It was a hook to allow for custom argv parsing.

zachallaun (approved):

zachdaniel:

Issue #177, opened 2024-12-27
Check if files have been moved


bradleygolden:
**Is your feature request related to a problem? Please describe.**

First, I want to say I love the `mix igniter.move_files` task and `igniter` in general. 😊 The issue is that after running this function on my codebase, over time I occasionally make mistakes by placing files in the wrong location without realizing it. Without any kind of continuous check, there’s no way to catch these errors unless I manually run the command again.

**Describe the solution you'd like**

It would be ideal to have a check feature to ensure files are in their correct locations without moving them automatically. I could run this locally alongside my tests and in CI.

**Describe alternatives you've considered**

- I could run the task periodically to verify that everything is in place, but I find this workflow inconvenient, especially when working on a team with other developers.

- I could also add a custom mix task to my codebase to perform this check, but I feel it would make more sense for this functionality to be part of `igniter`.

**Express the feature either with a change to resource syntax or with a change to the resource interface**

I imagine this feature could work similarly to `mix format --check-formatted`. For example, `mix igniter.move_files --check` or `mix igniter.move_files --check-not-moved`.

**Additional context**

I already have a proof of concept [here](https://github.com/albert-io/igniter/commit/6e42c686d4d1b2d78c2514afbb0ef7227c092e28), but I’m waiting for 👍 before opening a PR to avoid unnecessary noise.

zachdaniel:
All igniter tasks supports the `--check` option. Does that not work?

bradleygolden:
Wow, I missed that somewhere. Did a quick check locally and it looks like it works. 🙏

I'll let you know if I run into any issues.

bradleygolden:
Ahh actually it doesn't work when there are no files to move.

```
vscode ➜ /workspace/app (12-27-move_files) $ mix igniter.move_files --check
Compiling 101 files (.ex)
Compiling 11 files (.ex)
Generated myapp app
Finding all modules and determining proper locations...
** (CondClauseError) no cond clause evaluated to a truthy value
    (igniter 0.5.0) lib/igniter.ex:1099: Igniter.halt_if_fails_check!/3
    (igniter 0.5.0) lib/igniter.ex:973: Igniter.do_or_dry_run/2
    (mix 1.17.1) lib/mix/task.ex:495: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.17.1) lib/mix/cli.ex:96: Mix.CLI.run_task/2
[...]

PR #178, opened 2024-12-27
fix: allow --check to pass when no issues found

bradleygolden:
### Contributor checklist

Fixes https://github.com/ash-project/igniter/issues/177#issuecomment-2563939958

Is it necessary to regression test this? If so, what would be the best way? I had a problem reproducing the exact same issue in tests but it's easy to reproduce by creating a new project, installing igniter and running `mix igniter.move_files --check`.


zachdaniel (approved):

zachdaniel:

it’s straightforward so no worries :)

PR #185, opened 2025-01-03
improvement: print version diff when upgrading packages

adamtharani:
## Implements

- Prints out the version diff when upgrading a package; similar behaviour to what `mix deps.upgrade` does

Please let me know if a test is wanted for this 

### Contributor checklist


zachdaniel:

PR #186, opened 2025-01-05
improvement: capture and suppress output in installers

zachdaniel:
### Contributor checklist


zachdaniel:
Some frustrating realities:

I had to copy the same code into the installer as well so that it could do the same loading behavior. We will have to keep that code in sync as we do with the `igniter.install` task. We can maybe get clever doing something in the compile process, but with the `igniter.install` task there are subtle differences. For now I've opted to just keep them separate.

PR #190, opened 2025-01-13
fix: `Igniter.Project.MixProject.update/4` now creates non-existing functions

zachallaun:
This was an oversight when I originally implemented this: if the given top-level function doesn't exist, `update/4` is supposed to create it. (That behavior is explicitly stated in the docstring, actually.)


zachdaniel (approved):

zachdaniel:

PR #207, opened 2025-01-24
feat: `Igniter.Code.Module.move_to_attribute_definition`

leandrocp:


zachdaniel:
I think I'd rather have this in `Igniter.Code.Module.move_to_attribute_definition/2`. What do you think?

leandrocp:
> I think I'd rather have this in `Igniter.Code.Module.move_to_attribute_definition/2`. What do you think?

Makes way more sense. I'm gonna move it.

leandrocp:
@zachdaniel is jumping between attributes expected? for eg:

```elixir
 {:ok, zipper} = Igniter.Code.Module.move_to_attribute_definition(zipper, :doc)
 # do something with @doc

# then continue moving to another attribute
{:ok, zipper} = Igniter.Code.Module.move_to_attribute_definition(zipper, :foo_key)
 ```

zachdaniel:
It is unlikely that you would chain them like that. You'd likely start at the "top", in case they aren't in the order you expect.

leandrocp:
> It is unlikely that you would chain them like that. You'd likely start at the "top", in case they aren't in the order you expect.

Gotcha then I think the current implementation is correct.

zachdaniel (approved):

zachdaniel:

PR #212, opened 2025-01-30
Long diff handling

zachdaniel:
### Contributor checklist


zachdaniel:
Actually never mind. We've found that the generating of the diff is long enough that this will add a lot of delay :)

PR #213, opened 2025-01-30
Add `:after` opt to `Config` functions

leandrocp:
Expose an opt to move to a matching node.

* Add `:after` opt to `Config` functions
    - `configure_new/6`
    - `configure/6`
    - `modify_configuration_code/5`
    
* Replace `updater` arg with `opts` in `modify_configuration_code/5`
This introduces a breaking change but it seems more appropriate than adding a new arg. I can revert it tho.

---

Notes:
1. I can push more docs and examples, I just wanted to confirm this looks good first.
2. `move_to/2` is recursive already

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -7,10 +7,10 @@ defmodule Igniter.Code.Common do
   alias Sourceror.Zipper
 
   @doc """
-  Moves to the next node that matches the predicate.
+  Moves to the first node that matches the predicate.
   """
   @spec move_to(Zipper.t(), (Zipper.t() -> boolean())) :: {:ok, Zipper.t()} | :error
-  def move_to(zipper, pred) do
+  def move_to(zipper, pred) when is_function(pred, 1) do
     if pred.(zipper) do
       {:ok, zipper}
     else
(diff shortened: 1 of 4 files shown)


zachdaniel:
Move to is recursive insofar as it keeps going until the first match. We want to keep going to the *last* match. At least, I think that makes more sense for an `after` option?

We should add a pattern match for `updater` being a function, and should convert it to opts automatically and `IO.warn` that the call is deprecated.

leandrocp:
> Move to is recursive insofar as it keeps going until the first match. We want to keep going to the last match. At least, I think that makes more sense for an after option?

Hi @zachdaniel do you have an example where it would need to go to the last match?

For the example here https://github.com/ash-project/igniter/pull/213/files#diff-3cc69527888feb8bfded5d1bfcbbaf27416b6f65765315669ed39cdb7b9156bfR416 `move_to` returns the exact node I want, ie: `{:=, _, _}` so calling it again falls into an infinite loop. The current code works (I think) because later in the code it calls `Zipper.right()`.

I've updated the tests to make it more clear.

zachdaniel:
> For the example in tests, move_to returns the exact node I want, ie: {:=, _, _} so calling it again falls into an infinite loop. The current code works (I think) because later in the code it calls Zipper.right().

Right, so we should add in a `next` before checking. The basic idea is that the pattern you provide could appear more than once, and you have no control over where in the file it appears. So the `after` option *could* mean "after first match" or it could mean "after last match". In other places (like adding children to supervisors) we use `after` to mean "after last match". I think it should mean that in this case too.

For instance, taking your example, if the config had:

```elixir
port = 1000

port = 2000
```

where would you want it to go? There isn't a *great* answer, but I'd suggest that after the second `port = ` would make the most sense :)

leandrocp:
> I'd suggest that after the second port =  would make the most sense

In that case, I've used `Sourceror.traverse/3` since it needs to traverse the whole tree. The only difference is that function will only traverse the subtree if the zipper is not a the top, which seems okay? If for some reason you pass a modified `zipper` than that would be expected, right?

lib/igniter/code/common.ex
```diff
@@ -24,6 +24,36 @@ defmodule Igniter.Code.Common do
...
+  ```elixir
+  port = 4000
+  port = 4001
+  ```
+
+  With a match for `port = _` as `{:=, _, [{:port, _, _}, _]}`,
+  will return the second `port` variable.
+  """
+  @spec move_to_last(Zipper.t(), (Zipper.t() -> boolean())) :: {:ok, Zipper.t()} | :error
+  def move_to_last(zipper, pred) when is_function(pred, 1) do
```
leandrocp, line 42: Added this function since this pattern seems common enough to be exposed. I have other use cases where an `:after` option would be useful and use `move_to_last/2` too.

zachdaniel (approved):

zachdaniel:
Yes, agreed. Everything with zippers should start from wherever the current position is (in terms of our interfaces)

zachdaniel:

PR #220, opened 2025-02-06
Change default updater fn for configure_runtime_env/6 to match configure/6

### LINKED ISSUE #216: Config `:updater` option default value inconsistency
I was looking at the default behavior of various `Igniter.Project.Config` functions when the `:updater` option is not provided by the user.

In `configure/6` this default is to overwrite the existing value:
https://github.com/ash-project/igniter/blob/v0.5.21/lib/igniter/project/config.ex#L172

However, in `configure_runtime_env/6` the default is a no-op:
https://github.com/ash-project/igniter/blob/v0.5.21/lib/igniter/project/config.ex#L83
and
https://github.com/ash-project/igniter/blob/v0.5.21/lib/igniter/project/config.ex#L100

I stumbled upon this inconsistency earlier today because I thought both would overwrite by default.  Maybe I overlooked something...

APB9785:
## Resolves #216

### DIFF
--- a/lib/igniter/project/config.ex
+++ b/lib/igniter/project/config.ex
@@ -80,7 +80,8 @@ defmodule Igniter.Project.Config do
             config_path,
             app_name,
             modify_to,
-            updater: opts[:updater] || (&{:ok, &1})
+            updater:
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #221, opened 2025-02-06
fix: `web_module/1` duplicating Web suffix

### LINKED ISSUE #219: `Igniter.Libs.Phoenix.web_module/1` returns wrong name in Umbrella apps
I noticed Igniter can't find routers in umbrella apps due to generating wrong web module names. Take for instance an umbrella project generated as:

```sh
mix phx.new.web cms_admin_web
```

`Igniter.Libs.Phoenix.web_module/1` returns `CmsAdminWebWeb` (duplicate `Web`).

I'm aware there's no reliable way to find the web module name but an idea to fix it is to traverse the existing module looking for `use MOD, :router`, ie: a web-ish module using `:router` which seems to be reliable enough to determine the web module prefix, otherwise if not found fallbacks to `project prefix <> "Web"`.

Wdyt?

leandrocp:
Close #219

### DIFF
--- a/lib/igniter/libs/phoenix.ex
+++ b/lib/igniter/libs/phoenix.ex
@@ -7,7 +7,7 @@ defmodule Igniter.Libs.Phoenix do
   @spec web_module_name() :: module()
   @deprecated "Use `web_module/0` instead."
   def web_module_name do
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #232, opened 2025-02-20
Add support for passing options to an install

lawik:
Install can now be done with:

```
mix igniter.install nerves#runtime=false`
```

Producing:

```
{:nerves, "~> x.y.z", [runtime: false]}
```

To avoid a bunch of fun problems with spaces, commas and such the format is a query string. Spaces and other nasty characters will need to be encoded. The encoding exists in the standard library. It will allow creating Igniter install tasks for Nerves systems and similar that have some fairly particular options.

### Contributor checklist

### DIFF
--- a/lib/igniter/project/deps.ex
+++ b/lib/igniter/project/deps.ex
@@ -361,9 +361,18 @@ defmodule Igniter.Project.Deps do
       end
 
     quoted =
-      if opts[:dep_opts] do
+      if opts[:dep_opts] && opts[:dep_opts] != [] do
+        dep_opts =
+          Enum.map(opts[:dep_opts], fn
+            {key, {:code, val}} ->
+              {key, Sourceror.parse_string!(val)}
+
+            {key, value} ->
+              {key, value}
+          end)
+
         quote do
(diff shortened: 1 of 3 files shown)


zachdaniel:
So I don't mind supporting this syntax, but at the same time this feels like something that should be configured in the `%Info{}` of the installer that you're running, that way users don't have to know about it. Thoughts?

zachdaniel:
Although bringing in arbitrary values as code also activates my spider sense 😆

zachdaniel:
Right, but in most scenarios it would be super wild to try to actually type them in. I.e we support powershell and have already had issues along those lines 😆 See stuff like this: https://www.reddit.com/r/PowerShell/comments/ho4xei/quoting_argument_that_contains_quotes_for_a/

So you couldn't even necessarily reliably give people a command they can paste into their terminal to run.

So my point is that it might be so cumbersome/ridiculous to actually write these options that it's not worth making it a feature. Whereas allowing the installer to say `dep_opts: [runtime: false]` would solve your immediate issue and has no such usability concerns. So maybe we start there?

zachdaniel:
Done: c0a7fec03e58d3d6e4474f0d6deb485639477f7c

PR #236, opened 2025-02-21
fix: find app_name when kw list is defined at the end

### LINKED ISSUE #235: Igniter expects the definition body of Mix.Project#project to be exactly a keywordlist.
**Describe the bug**
In a mix project. If the content of the project function is not a single list, then Igniter can't find the project app and returns an error suggesting that it isn't set. 

```
compiling beacon, beacon_live_admin ✔
** (RuntimeError) Failed to parse the application name from mix.exs.

Please ensure that the `project` function in your mix.exs
file returns a keyword list with an `app` key.

    (igniter 0.5.27) lib/igniter/project/application.ex:36: anonymous fn/1 in Igniter.Project.Application.app_name/1
    (beacon 0.4.1) lib/mix/tasks/beacon.install.ex:93: 
```

**To Reproduce**
In a mix project:
```
def project do
  version = "0.0.0"
  [
      app: :my_app,
      version: version,
      aliases: aliases(),
      deps: deps()
  ]
end
```

**Expected behavior**
I didn't expect that anything would care about anything other than the return value of a function.
Either Igniter should look for the last list in the project function definition, 
or - ideally - the return value of the function, 
or at least have an error message that states that it expects the body of the project function definition to only be a keyword list.

\*\* Runtime

- Elixir version `1.17.3`
- Erlang version `OTP 27`
- Igniter version `0.5.27`

**Additional context**
Encountered trying to install BeaconCMS

leandrocp:
Close #235

So the logic here is to find where `:app` is defined inside `project/0`, which does cover the case described at #235 and also cases where the kw list is assigned to a variable. Although it wouldn't cover some edge cases like for eg:

```elixir
def project do
  default = [
    app: :igniter_test
  ]

  Keyword.merge(default, other())
end
```

But I'm not sure if that should even be supported. Let me know if there's a better approach.

### DIFF
--- a/lib/igniter/project/application.ex
+++ b/lib/igniter/project/application.ex
@@ -26,8 +26,10 @@ defmodule Igniter.Project.Application do
       |> Sourceror.Zipper.zip()
 
     with {:ok, zipper} <- Igniter.Code.Function.move_to_def(zipper, :project, 0),
-         zipper <- Igniter.Code.Common.rightmost(zipper),
-         true <- Igniter.Code.List.list?(zipper),
+         {:ok, zipper} <-
+           Igniter.Code.Common.move_to(zipper, &match?({:__block__, _, [:app]}, &1.node)),
+         zipper = Zipper.up(zipper),
+         zipper = Zipper.up(zipper),
          {:ok, zipper} <- Igniter.Code.Keyword.get_key(zipper, :app),
          {:ok, app_name} when is_atom(app_name) <- expand_key_value(zipper) do
       app_name
--- a/test/igniter/project/application_test.exs
+++ b/test/igniter/project/application_test.exs
@@ -231,6 +231,56 @@ defmodule Igniter.Project.ApplicationTest do
       assert Igniter.Project.Application.app_name(igniter) == :igniter_test
     end
 
+    test "it returns the application name when keyword list is defined at the end" do
+      igniter =
+        test_project(
+          files: %{
+            "mix.exs" => """
+            defmodule IgniterTest.MixProject do
+              use Mix.Project
+
+              def project do
+                releases = [
+                  test: []
+                ]
+
+                [
+                  app: :igniter_test,
+                  releases: releases
+                ]
+              end
+            end
+            """
+          }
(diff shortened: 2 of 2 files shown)


lib/igniter/project/application.ex
```diff
@@ -26,8 +26,10 @@ defmodule Igniter.Project.Application do
       |> Sourceror.Zipper.zip()
 
     with {:ok, zipper} <- Igniter.Code.Function.move_to_def(zipper, :project, 0),
-         zipper <- Igniter.Code.Common.rightmost(zipper),
-         true <- Igniter.Code.List.list?(zipper),
```
leandrocp, line 30: `Keyword.get_key/2` does check if it's a list already.

lib/igniter/project/application.ex
```diff
@@ -26,8 +26,10 @@ defmodule Igniter.Project.Application do
...
+         {:ok, zipper} <-
+           Igniter.Code.Common.move_to(zipper, &match?({:__block__, _, [:app]}, &1.node)),
```
zachdaniel, line 30: I think this might be too liberal. Any other option contained that contains an `app` option, anywhere in the structure could be construed as the app config we are looking for, no matter how deep we search. I think we should start simpler, going to the rightmost literal list, and otherwise complaining. We have a `deps_location` pattern for allowing users to configure where in the deps function their list is, so we could eventually do similar patterns here, but this method can yield false *positives* which I think is more dangerous.

zachdaniel: I was busy when I typed the above, sorry 😆 its not very clear.

The current feature could have a false positive, i.e if someone had something like this:

```elixir
def project() do
  [
    foo: [app: :foo], # <- we'd find this one.
   app: :foo
  ]
end
```

And false positives are dangerous. I think "left most list" is really the only safe heuristic to apply for now.

leandrocp: It was clear enough 😄 

Yeah that makes sense, I'm gonna look at that `deps_location` and soon push changes to this branch.

PR #248, opened 2025-03-08
Switch to doctest for Igniter.Code.Common.add_code/3

axelson:
Fixes a typo in the existing docs and switches to doctests to ensure that the docs do not get out of date.

But I am not quite 100% convinced that this is necessarily more readable, it feels a bit awkward to me. I wish I could write a better assertion as the last line of the doctest, we could use a heredoc but then we'd have to annoyingly deal with the trailing newline.

Here's the new and old docs:
| Old       | New     |
| ---------- | ------ |
| Screenshot 2025-03-08 14-17-05@2x | Screenshot 2025-03-08 14-17-27@2x |

### Contributor checklist

### DIFF
--- a/lib/igniter/code/common.ex
+++ b/lib/igniter/code/common.ex
@@ -257,31 +257,18 @@ defmodule Igniter.Code.Common do
 
   ## Example:
 
-  ```elixir
-  existing_zipper = \"\"\"
-  IO.inspect("Hello, world!")
-  \"\"\"
-  |> Sourceror.parse_string!()
-  |> Sourceror.Zipper.zip()
-
-  new_code = \"\"\"
-  IO.inspect("Goodbye, world!")
-  \"\"\"
-
-  existing_zipper
-  |> Igniter.Common.add_code(new_code)
-  |> Sourceror.Zipper.root()
-  |> Sourceror.to_string()
-  ```
-
-  Which will produce
-
-  ```elixir
-  \"\"\"
-  IO.inspect("Hello, world!")
-  IO.inspect("Goodbye, world!")
-  \"\"\"
-  ```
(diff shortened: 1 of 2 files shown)


zachdaniel:
Hmm...yeah that's interesting. The value of the secrets is pretty high even if it looks a bit less good. When you click the button to copy this do you get all the `...>` also? Maybe we can clean it up by reworking the example to be single line strings in places?

axelson:
> When you click the button to copy this do you get all the ...> also? Maybe we can clean it up by reworking the example to be single line strings in places?

Clicking the button (or selecting the text directly) doesn't include the `...>`. If you generate the docs with my makeup_elixir PR https://github.com/elixir-makeup/makeup_elixir/pull/39 then you get:
```
existing_zipper = Igniter.Code.Common.parse_to_zipper!("""
IO.inspect("abc")
""")

existing_zipper
|> Igniter.Code.Common.add_code("""
IO.inspect("Goodbye, world!")
""", after: true)
|> Sourceror.Zipper.root()
|> Sourceror.to_string()
"IO.inspect(\"abc\")
IO.inspect(\"Goodbye, world!\")"
```
which can be directly copied and pasted to an IEx terminal that has Igniter as a dep.

But yeah it isn't really the most readable. Maybe it would make more sense to split it into two, a smaller more limited doctest and a larger by hand example (I think we could make that larger example flow a bit better than the current version on `main`).

axelson:
> The value of the secrets is pretty high even...

Do you mean doctest instead of secrets?

zachdaniel:
Yes, that was autocorrect 😂

zachdaniel:
I'm down with this. It is slightly less readable but more conventional.

zachdaniel:
Good to merge?

axelson:
Yeah, let's do it! 🎉

PR #249, opened 2025-03-08
Improve the Readme

axelson:
Rewrite portions of the readme to help understandability
Add an "audiences" section to directly mention and speak to the two main audiences of Igniter
Split the readme into end-user and library author sections
- Note: I put the library author section later because I assume that end-users are more common and library authors are a bit more dedicated/willing to scroll to find the content that is specific to them (plus they are end-users as well)

Other changes:
- Consistently use Igniter with a capital I
- Make some wording more direct instead of passive
- Consistently use `end-user`

The updated readme can be viewed at: https://github.com/axelson/igniter/blob/improve-docs/README.md

### Contributor checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,11 @@
 Igniter is a code generation and project patching framework.
 
 There are two audiences for Igniter:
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):

zachdaniel:

PR #251, opened 2025-03-10
Add `:placement` option to `Phoenix.add_scope/4` and `Phoenix.append_to_scope/4`

leandrocp:
Hey @zachdaniel here's a PoC to allow customization of routes placement. To fix that conflict between `beacon_site "/"` and Ash Admin in the ash-hq.org generator and also to fix https://github.com/BeaconCMS/beacon/issues/744

I'm not sure about the names in the `:placement` option (Idk if there's a pattern) but `:before_first_scope` is the current behavior that tries to place the route before the first scope, or after the last pipeline, or after use Router. The new option `: bottom_of_module` (not a great name) just place the code at the end of the file.

Here's how the patched file looks like:

```elixir
defmodule TestWeb.Router do
  use TestWeb, :router

  use Beacon.Router

  pipeline :beacon do
    plug Beacon.Plug
  end

  pipeline :browser do
    plug(:accepts, ["html"])
    plug(:fetch_session)
    plug(:fetch_live_flash)
    plug(:put_root_layout, html: {TestWeb.Layouts, :root})
    plug(:protect_from_forgery)
    plug(:put_secure_browser_headers)
  end

  pipeline :api do
    plug(:accepts, ["json"])
  end

  scope "/", TestWeb do
    pipe_through(:browser)

    get("/", PageController, :home)
  end

  # Other scopes may use custom stacks.
  # scope "/api", TestWeb do
  #   pipe_through :api
  # end

  # Enable LiveDashboard and Swoosh mailbox preview in development
  if Application.compile_env(:test, :dev_routes) do
    # If you want to use the LiveDashboard in production, you should put
    # it behind authentication and allow only admins to access it.
    # If your application does not have an admins-only section yet,
    # you can use Plug.BasicAuth to set up some basic authentication
    # as long as you are also using SSL (which you should anyway).
    import Phoenix.LiveDashboard.Router

    scope "/dev" do
      pipe_through(:browser)

      live_dashboard("/dashboard", metrics: TestWeb.Telemetry)
      forward("/mailbox", Plug.Swoosh.MailboxPreview)
    end
  end

  scope "/", alias: TestWeb do
    pipe_through [:browser, :beacon]
    beacon_site "/", site: :my_site
  end
end


```

### DIFF
--- a/lib/igniter/libs/phoenix.ex
+++ b/lib/igniter/libs/phoenix.ex
@@ -135,9 +135,7 @@ defmodule Igniter.Libs.Phoenix do
 
   - `:router` - The router module to append to. Will be looked up if not provided.
   - `:arg2` - The second argument to the scope macro. Must be a value (typically a module).
-  - `:placement` - `:before` | `:after`. Determines where the `contents` will be placed. Note that it first tries
-    to find a matching scope and place the contents into that scope, otherwise `:placement` is used to determine
-    where to place the contents:
+  - `:placement` - `:before` | `:after`. Determines where the `contents` will be placed:
     - `:before` - place before the first scope in the module if one is found, otherwise tries to place
        after the last pipeline or after the `use MyAppWeb, :router` call.
     - `:after` - place at the end (bottom) of the module, after all scopes and pipelines.
--- a/lib/igniter/libs/phoenix.ex
+++ b/lib/igniter/libs/phoenix.ex
@@ -199,6 +199,8 @@ defmodule Igniter.Libs.Phoenix do
   * `:router` - The router module to append to. Will be looked up if not provided.
   * `:arg2` - The second argument to the scope macro. Must be a value (typically a module).
   * `:with_pipelines` - A list of pipelines that the pipeline must be using to be considered a match.
+  # TODO: improve docs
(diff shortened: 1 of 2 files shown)


leandrocp:
Note that it doesn't handle the default Phoenix or LiveView route, that's a separate concern I have to handle in Beacon.

lib/igniter/libs/phoenix.ex
```diff
@@ -199,6 +199,8 @@ defmodule Igniter.Libs.Phoenix do
   * `:router` - The router module to append to. Will be looked up if not provided.
   * `:arg2` - The second argument to the scope macro. Must be a value (typically a module).
   * `:with_pipelines` - A list of pipelines that the pipeline must be using to be considered a match.
+  # TODO: improve docs
+  * `:placement` - :before_first_scope | :bottom_of_module - defaults to :before_first_scope (current behavior)
```
zachdaniel, line 203: I think maybe it could be as simple as `:first` or `:last`, and we can evolve from there. We can't guarantee much past that anyway since the router could have all kinds of stuff in it. But we can at least guarantee that it goes after all other routes & scopes or before all other routes and scopes.

zachdaniel:
LGTM, some commentary on the naming but otherwise 👍

zachdaniel (approved):

zachdaniel:

PR #257, opened 2025-03-20
fix: igniter.upgrade crash on dependency when only option is an atom

stentroad:
### Contributor checklist


zachdaniel:

PR #265, opened 2025-04-05
docs: add spec for within

ken-kost:
### Contributor checklist


ken-kost:
Okay so I'm not sure what a good simple example would be for within but I could write up a guide that shows you how to modify module config, one function and also add another function to that module. I think that would be a verbose enough example to show the functionality of within. wdyt?

zachdaniel (approved):

zachdaniel:

PR #273, opened 2025-05-10
Matches function guards when using move_to_def

danwanek-wm:
### Contributor checklist

I noticed when trying to use `Function.move_to_def` that I was getting an `:error` when the function hard a guard clause. This should fix that issue as well as test for it.

Definitely let me know if this isn't idiomatic to the way you want the code in Igniter. Happy to work with whatever.

Cheers,

Dan


zachdaniel:

PR #274, opened 2025-05-11
Task/adds move to function and attrs

danwanek-wm:
### Contributor checklist

This adds a function that will put you at the top of the function and will take into consideration common attributes. So when you do a `Common.add_code` with a `placement: :before`, it will work nicely. I also added a `@doc` that shows how to do this.

Thanks again for this great library @zachdaniel. It's been a pleasure to work with. It took me some time to wrap my mind around Sourceror, but the layout is really clean and easy to follow. If you're ever passing through Black Mountain, let me know and I'll buy you a beer. ✌️

### DIFF
--- a/lib/igniter/code/function.ex
+++ b/lib/igniter/code/function.ex
@@ -52,7 +52,7 @@ defmodule Igniter.Code.Function do
     end
   end
 
-  defp do_move_to_def(zipper, fun, arity, kind) do
+  defp do_move_to_def(zipper, fun, arity, kind, move_to_do \\ true) do
     case Common.move_to(zipper, fn zipper ->
            case Zipper.node(zipper) do
              # Match the standard function definition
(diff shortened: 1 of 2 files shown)


zachdaniel:
This looks great, I'm thinking that we should make this an option to the `move_to_def` function, what do you think?

danwanek-wm:
> This looks great, I'm thinking that we should make this an option to the `move_to_def` function, what do you think?

I like that idea. It minimizes the function footprint and they seem pretty related. What do you think about adding the option as `destination: [:before | :in | :after]` or do you think `:placement` makes more sense to be consistent with the `add_code` function?

I'll try and take a stab at this soon. I'm traveling out of town this week for work so I'm not sure how much time I'll have until the following week

zachdaniel:
Hmm.......... yeah probably not `:placement` as thats for adding code. I think `target` might be better, and we can use that pattern elsewhere as needed.

danwanek-wm:
@zachdaniel  - I was able to find some time and took a stab at adding it as an option. LMK what you think and if you would like it changed in any way. ✌️

lib/igniter/code/function.ex
```diff
@@ -24,35 +25,98 @@ defmodule Igniter.Code.Function do
     end
   end
 
-  @spec move_to_defp(Zipper.t(), fun :: atom, arity :: integer | list(integer)) ::
+  @spec move_to_defp(Zipper.t(), fun :: atom, arity :: integer | list(integer), [opts]) ::
```
zachdaniel, line 28:
```suggestion
  @spec move_to_defp(Zipper.t(), fun :: atom, arity :: integer | list(integer), Keyword.t()) ::
```

lib/igniter/code/function.ex
```diff
@@ -24,35 +25,98 @@ defmodule Igniter.Code.Function do
...
-    do_move_to_def(zipper, fun, arity, :defp)
+        when opts: {:target, :inside | :before | :at}
+  def move_to_defp(zipper, fun, arity, opts \\ [target: :inside]) do
+    do_move_to_def(zipper, fun, arity, :defp, opts)
   end
 
-  @spec move_to_def(Zipper.t()) :: {:ok, Zipper.t()} | :error
-  def move_to_def(zipper) do
-    move_to_function_call(zipper, :def, :any)
+  @spec move_to_def(Zipper.t(), [opts]) :: {:ok, Zipper.t()} | :error
```
zachdaniel, line 35:
```suggestion
  @spec move_to_def(Zipper.t(), Keyword.t()) :: {:ok, Zipper.t()} | :error
```

lib/igniter/code/function.ex
```diff
@@ -24,35 +25,98 @@ defmodule Igniter.Code.Function do
...
+  #
+  #   @doc "hello"
+  #   @spec hello() :: :world
+  #   def hello() do
+  #     :world
+  #   end
+  # end
+  ```
+  """
+  @spec move_to_def(Zipper.t(), fun :: atom, arity :: integer | list(integer), [opts]) ::
```
zachdaniel, line 101:
```suggestion
  @spec move_to_def(Zipper.t(), fun :: atom, arity :: integer | list(integer), Keyword.t()) ::
```

lib/igniter/code/function.ex
```diff
@@ -24,35 +25,98 @@ defmodule Igniter.Code.Function do
...
+  defp do_move_to_def(zipper, fun, [arity | rest], kind, opts) do
+    case do_move_to_def(zipper, fun, arity, kind, opts) do
       {:ok, zipper} -> {:ok, zipper}
-      :error -> do_move_to_def(zipper, fun, rest, kind)
+      :error -> do_move_to_def(zipper, fun, rest, kind, opts)
     end
   end
 
-  defp do_move_to_def(zipper, fun, arity, kind) do
+  defp do_move_to_def(zipper, fun, arity, kind, target: target) do
```
zachdaniel, line 119: we shouldn't pattern match on keywords (because they can be in any order)

lib/igniter/code/function.ex
```diff
@@ -24,35 +25,98 @@ defmodule Igniter.Code.Function do
...
+  #   end
+  # end
+  ```
+  """
+  @spec move_to_def(Zipper.t(), fun :: atom, arity :: integer | list(integer), Keyword.t()) ::
           {:ok, Zipper.t()} | :error
-  def move_to_def(zipper, fun, arity) do
-    do_move_to_def(zipper, fun, arity, :def)
+        when opts: {:target, :inside | :before | :at}
+  def move_to_def(zipper, fun, arity, opts \\ [target: :inside]) do
```
zachdaniel, line 104: instead of defaulting the `opts` to a value, we should default to an empty list and use tools like `Keyword.get(opts, :target, :inside)` to access the default value.

zachdaniel (approved):

zachdaniel:

PR #283, opened 2025-05-17
Add igniter.init task to igniter_new archive

lawik:
It adds itself as an optional dependency with `runtime: false`

It then uses Igniter to generate an install task.

### Contributor checklist


lawik:
I will admit that I don't get all what the archive igniter.install task does so might be some stuff that could be removed here.

This should be a good start to initializing a library. Let me know if you want any changes :)

zachdaniel:
Let's call it `init_library` and then it looks good to me 👌 want to make sure it doesn't confuse end users as it would cause them to ship mix to prod if they ran this.

Issue #287, opened 2025-05-21
igniter cannot be compiled with OTP/28 because of the upstream break


vkryukov:
When compiling igniter with OPT/28 and Elixir 1.18.4, it breaks because OTP/28 does not allow regexs in module constants anymore, which are extensively used by inflex. I created an issue upstream as well: https://github.com/nurugger07/inflex/issues/98, but since that project is not very frequently updated, I'm adding a bug here to track the issue.

gsarwate:
+1

Error details:


== Compilation error in file lib/inflex/pluralize.ex ==

** (ArgumentError) cannot inject attribute @singular into function/macro because cannot escape #Reference<0.515019536.430309385.174081>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity
    
    (elixir 1.18.4) lib/kernel.ex:3777: Kernel.do_at_escape/2
    (elixir 1.18.4) expanding macro: Kernel.@/1
    lib/inflex/pluralize.ex:179: Inflex.Pluralize.singularize/1
could not compile dependency :inflex, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile inflex --force", update it with "mix deps.update inflex" or clean it with "mix deps.clean inflex"

Igniter dependency:

      {:inflex, "~> 2.0"}

vkryukov:
Should be fixed by #288

PR #288, opened 2025-05-21
Fix issue #287

vkryukov:
- Fix configuration handling in IgniterConfig to be compatible with OTP/28
- Import Inflex.Pluralize from inflex module with OTP/28 compatibility fixes
- Remove :inflex and :mix_test_watch from compatibility list


zachdaniel:
Mind unlocking the unused deps? See the failing check for more 🙇

vkryukov:
> Mind unlocking the unused deps? See the failing check for more 🙇

Yep, just saw that and pushed a fix.

zachdaniel:

zachdaniel:
Hmm...I was expecting this not to cause a problem for packages not using `inflex`. https://github.com/ash-project/igniter/actions/runs/15172174136/job/42664647523

zachdaniel:
Alright, we unfortunately have to roll this back, as it actually does break other folks installers.

vkryukov:
Interesting. I assume those actions are trying to compile some libraries (e.g., ash) that are dependent on igniter?

At least with ash, if I manually replace `igniter` import in `mix.exs` with `{:igniter, path: "../igniter", optional: true, override: true}` in a fresh clone of the ash repo on my machine, it compiles and runs `mix test --only igniter` without any problems. Without `override: true`, it complains about a conflict.

zachdaniel:
Hmm...weird. I'm doing that with oban now and I'm getting the error shown in tests.

vkryukov:
This is certainly weird. I repeated the same process with oban-bg/oban:
- git clone the repo
- replace the igniter import with the path to local igniter as per above (and double-checked that I'm on my branch)
- using the same tools (elixir 1.18.1, erlang 27.2)
- mix deps.get
- mix test --only igniter

runs just fine.

zachdaniel:
Its all good, we're not sure why it happens, but we're rolling forward with your changes 🙇

vkryukov:
Great! 'Cause I has started to worry about the color of my nickname changing back from pink to whatever color it was before ;)

zachdaniel:
😆 its a one-way road, but yes all good here 😆

PR #291, opened 2025-05-29
feat: add assert_creates_file/3 macro to Igniter.Test

lawik:
Add a new testing macro that allows users to assert that a file was created and test its contents using a function. This provides a more flexible API for testing file creation compared to the existing assert_creates/3 macro.

The new macro supports:
- Pipeline-style testing with |> assert_creates_file(path, fn file -> ... end)
- Custom assertions on file content using pattern matching and regex
- Proper error handling and stack trace preservation

### Contributor checklist

### DIFF
--- a/lib/igniter/test.ex
+++ b/lib/igniter/test.ex
@@ -623,7 +623,11 @@ defmodule Igniter.Test do
   @doc """
   Asserts that a file was created during the igniter run.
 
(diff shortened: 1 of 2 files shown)


lawik:
Tidying up old PRs. Feel free to close if this isn't relevant anymore but from a glance it would still be useful.

zachdaniel:

Issue #67, opened 2024-08-03
Add `Igniter.Project.Config.remove_configuration`


ibarakaiev:
**Is your feature request related to a problem? Please describe.**
Currently, it is possible to configure keys, but not possible to remove existing configured keys.

**Describe the solution you'd like**
`Config.remove_configuration` or `Config.remove_key` or something like that.

**Describe alternatives you've considered**
It's currently possible to run `Config.configure/6` and pass `nil` or `[]` or something like that to clear the configuration, but it would be ideal if there's a dedicated function to remove configuration instead, which should also remove the entire `config :key ...` line as well if everything became empty.

PR #309, opened 2025-06-21
Implement removal of configuration

LostKobrakai:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

Handles parts of #67

### DIFF
--- a/lib/igniter.ex
+++ b/lib/igniter.ex
@@ -504,21 +504,51 @@ defmodule Igniter do
   """
   @spec update_elixir_file(t(), Path.t(), zipper_updater()) :: Igniter.t()
   def update_elixir_file(igniter, path, func) do
-    if Rewrite.has_source?(igniter.rewrite, path) do
-      igniter
-      |> apply_func_with_zipper(path, func)
-      |> format(path)
-    else
-      if exists?(igniter, path) do
-        source = read_ex_source!(igniter, path)
-
-        %{igniter | rewrite: Rewrite.put!(igniter.rewrite, source)}
-        |> format(path)
+    case use_or_read_elixir_file(igniter, path) do
+      {:exists, igniter} ->
+        igniter
         |> apply_func_with_zipper(path, func)
         |> format(path)
-      else
+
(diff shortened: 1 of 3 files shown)


lib/igniter.ex
```diff
@@ -504,21 +504,51 @@ defmodule Igniter do
...
         add_issue(igniter, "Required #{path} but it did not exist")
-      end
+    end
+  end
+
+  @doc """
+  Updates the source code of the given elixir file if it exists
+  """
+  @spec update_existing_elixir_file(t(), Path.t(), zipper_updater(), keyword) :: Igniter.t()
+  def update_existing_elixir_file(igniter, path, func, opts \\ []) do
```
zachdaniel, line 522: could this instead be done via `update_elixir_file(...., required?: false)`? Then we wouldn't need another function.

lib/igniter/project/config.ex
```diff
@@ -294,6 +276,31 @@ defmodule Igniter.Project.Config do
     end)
   end
 
+  def remove_configuration(igniter, file_name, app_name) do
```
zachdaniel, line 279: We may want to give this a name that indicates that it effectively can only remove an "entire application's" configuration from a configuration, like `remove_app_configuration` or something like that. Additionally, the same application could be configured multiple times in this file, so this will need to recursively check again, only returning `:error` if it never found a match.

lib/igniter.ex
LostKobrakai: Probably. I didn't find the other related functions to be particularly consistent around this topic.

zachdaniel: What we have currently is an `on_exists` configuration for `create_new_file` for example, so perhaps we can have an `on_not_exists` configuration which defaults to `:error`.

LostKobrakai: There's also a `required?` option for `include_existing_file` with an `include_or_create_file` counterpart. I'm not particularly interested in any option over the other. I don't think there's much of a line right now though.

PR #332, opened 2025-09-22
test: modify_config_code twice with keyword values

grzuy:
# Contributor checklist

- [x] AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

---

Hey @zachdaniel :wave: 

I spotted this issue while working on https://github.com/mimiquate/tower/pull/166.

And I was able to reproduce it here with this test case. But to be honest I am not sure I am calling the right functions and with the correct expectations.

I debugged a bit and the failure lead me to a few changes around Aug 2024...
- https://github.com/ash-project/igniter/pull/70
- https://github.com/ash-project/igniter/pull/71
- https://github.com/ash-project/igniter/pull/77
- https://github.com/ash-project/igniter/commit/853f27532cd2e321a72789eb9ee3b772d9f8dfe6

Thanks in advance!

### DIFF
--- a/test/igniter/project/config_test.exs
+++ b/test/igniter/project/config_test.exs
@@ -699,6 +699,25 @@ defmodule Igniter.Project.ConfigTest do
 
       assert String.contains?(config, "config :fake, foo: [bar: true]")
     end
+
+    test "update existing config twice" do
+      zipper =
+        ~s"""
+        import Config
+        """
+        |> Sourceror.parse_string!()
+        |> Sourceror.Zipper.zip()
+
+      config =
+        zipper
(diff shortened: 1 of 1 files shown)


zachdaniel:
Would you want to include that fix here?

grzuy:
Thanks for the feedback.

Included extra commit with that change.

Running `mix test` locally I only get 1 failure which I also get when running on `main`, so it might be something else about my local setup.

Workflow here awaiting approval to run. We'll see :slightly_smiling_face:

PR #334, opened 2025-09-26
feat(ci): add `mishka_chelekom` project to test ci

shahryarjb:
This adds the `mishka_chelekom` project to the GitHub Actions test matrix in `test-projects.yml` for CI test.
Because all Mishka Chelekom project tasks are Igniter, i did not add `--only igniter` like the other

Thank you in advance

### DIFF
--- a/.github/workflows/test-projects.yml
+++ b/.github/workflows/test-projects.yml
@@ -41,7 +41,7 @@ jobs:
             {
               org: "mishka-group",
               name: "mishka_chelekom",
-              test-cmd: "mix test",
+              test-cmd: "mix test --only igniter",
             },
           ]
         otp: ["27.2"]


.github/workflows/test-projects.yml
```diff
@@ -39,6 +38,11 @@ jobs:
               name: "oban",
               test-cmd: "mix test --only igniter",
             },
+            {
+              org: "mishka-group",
+              name: "mishka_chelekom",
+              test-cmd: "mix test",
```
zachdaniel, line 44: I'd prefer to only run tests related to the igniter installers, can you add a `---only igniter` here and tag the relevant tests? Even if they are all igniter tests at the moment, the point is to make sure igniter's build doesn't fail just because some unrelated test fails in mishka.

shahryarjb: @zachdaniel Yes, I understand. now my master branch tests have igniter flag and i edited the `.github/workflows/test-projects.yml` and added `mix test --only igniter`
Thank you

Issue #347, opened 2025-10-26
`set_map_key` creates invalid nested map when setting a key


ollien:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

```
Erlang/OTP 27 [erts-15.1.2] [source] [64-bit] [smp:24:24] [ds:24:24:10] [async-threads:1] [jit:ns]

IEx 1.18.4 (compiled with Erlang/OTP 27)
```

Igniter commit: `bd2b582cab7316f4084749fc20cc9eda22e0c960`

### Operating system

Fedora 34

### Current Behavior

```ex
test "inserts value into map if not present" do
  {:ok, zipper} =
    "%{}"
    |> Code.Common.parse_to_zipper!()
    |> Code.Map.set_map_key(:foo, :bar, &flunk/1)

  assert "%{foo: :bar}" == zipper |> Zipper.node() |> Sourceror.to_string()
end
```

This test fails with

```
  1) test set_map_key/4 inserts value into map if not present (Igniter.Code.Map.Test)
     test/igniter/code/map_test.exs:10
     Assertion with == failed
     code:  assert "%{foo: :bar}" == zipper |> Zipper.node() |> Sourceror.to_string()
     left:  "%{foo: :bar}"
     right: "%{%{foo: :bar}}"
     stacktrace:
       test/igniter/code/map_test.exs:16: (test)
```

### Reproduction

See above

### Expected Behavior

I would expect this to produce a zipper of `%{foo: :bar}`, but instead it produces `%{%{foo: :bar}}`, which is not a valid map.

I thought about just submitting a PR for this, but I get the feeling I might be misunderstanding the intended use of the function.

zachdaniel:
Looks like a good old fashioned bug to me, PR welcome!

PR #348, opened 2025-10-28
Fix `put_in_map`/`set_map_key` not setting keys properly

ollien:
In attempting to solve #347, I ran into other problems that all felt related enough to fix at once

1. Both `put_in_map` and `set_map_key` would produce an invalid "nested" map when inserting into an empty map (`%{%{foo: bar}}`)
2. Both `put_in_map` and `set_map_key` would call the updater, take its updated value, and then return the zipper at the position of the value. This was inconsistent with the non updater case (i.e. inserting a new key would point to the map itself), so I made these consistent.

I recognize #2 is technically a breaking change, but I figured this couldn't have been intentionally inconsistent. If you would prefer a deprecation/new function, let me know.

Fixes https://github.com/ash-project/igniter/issues/347

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/igniter/code/map.ex
+++ b/lib/igniter/code/map.ex
@@ -36,7 +36,7 @@ defmodule Igniter.Code.Map do
           value = Common.use_aliases(value, zipper)
 
           {:ok,
-           Zipper.replace(
+           Common.replace_code(
              zipper,
              mappify([key | rest], value)
            )}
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #372, opened 2026-04-14
fix: support all mix dep formats in recursively_compose_schema

guitsaru:
## Summary

`Igniter.Util.Info.recursively_compose_schema/4` merges `installs` and `adds_deps` from composed tasks into the parent using `Keyword.merge/2`. That only works when every dep entry is shaped like `{name, value}` — any other supported mix dep form (e.g. `{:credo, "~> 1.7", only: [:dev, :test]}` or `{:my_dep, git: "..."}`) causes `Keyword.merge/2` to raise:

```
** (ArgumentError) expected a keyword list as the second argument,
   got: [{:boundary, "~> 0.10", [runtime: false]}]
```

The `dep` type at `lib/mix/task/info.ex:129-132` already documents all the accepted forms (`{name, version}`, `{name, opts}`, `{name, version, opts}`), and the rest of `Igniter.Util.Info` (`add_deps/3`, the installs normalization) handles them. The composition path just hadn't been updated to match.

### Repro

```elixir
defmodule Mix.Tasks.Parent do
  use Igniter.Mix.Task
  def info(_, _) do
    %Igniter.Mix.Task.Info{
      composes: ["child"],
      adds_deps: [{:boundary, "~> 0.10", runtime: false}]
    }
  end
  def igniter(i), do: i
end

defmodule Mix.Tasks.Child do
  use Igniter.Mix.Task
  def info(_, _), do: %Igniter.Mix.Task.Info{adds_deps: [{:cloak, "~> 1.1"}]}
  def igniter(i), do: i
end

# `mix igniter.install parent` raises the ArgumentError above.
```

### Fix

Replace the two `Keyword.merge/2` calls with a private `merge_deps/2` helper that dedupes by dep name (the first element of each entry) and leaves the original tuple shape untouched. The parent task's entry still wins on conflict, matching the previous `Keyword.merge/2` semantics.

### DIFF
--- a/lib/igniter/util/info.ex
+++ b/lib/igniter/util/info.ex
@@ -481,8 +481,8 @@ defmodule Igniter.Util.Info do
             alias_conflicts: alias_conflicts(schema, composing_schema, composing_task_name),
             composes: rest,
             extra_args?: schema.extra_args? || composing_schema.extra_args?,
-            installs: Keyword.merge(composing_schema.installs, schema.installs),
-            adds_deps: Keyword.merge(composing_schema.adds_deps, schema.adds_deps)
+            installs: merge_deps(composing_schema.installs, schema.installs),
+            adds_deps: merge_deps(composing_schema.adds_deps, schema.adds_deps)
         },
         argv,
         parent,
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #374, opened 2026-04-19
improvement: add rename_module/4 and mix igniter.refactor.rename_modu…

### LINKED ISSUE #373: Rename a module across a project similar to rename_function
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

mix igniter.refactor.rename_module would rename a module everywhere it appears in a project — updating defmodule, alias, ... and moving the files to the correct location. 

Motivation:

It doesn't matter if I have to do it myself or let an agent do refactoring work. For a human it is a lot of search and replace and an agent can easily take 20 Minutes in a smaller codebase and burns a lot of tokens. Helps both ;)

### Describe the solution you'd like

Proposed API

Mix task:

mix igniter.refactor.rename_module Foo.Bar Foo.Baz

See the working git repo below. If the feature itself is accepted, I can make a proper pull request. I tried to already keep some conventions I saw in the codebase.

### Describe alternatives you've considered

I couldn't find good alternatives. I only found rename_function what is not the same as renaming a module.

### Additional context

I built this as a standalone package because I wasn't sure if this will be accepted. It is what I use daily for my refactorings:
https://github.com/exfoundry/igniter_rename_module

The core logic lives in [lib/igniter_rename_module/refactors/rename.ex](https://github.com/exfoundry/igniter_rename_module/blob/main/lib/igniter_rename_module/refactors/rename.ex) and the mix task in [lib/mix/tasks/igniter.refactor.rename_module.ex](https://github.com/exfoundry/igniter_rename_module/blob/main/lib/mix/tasks/igniter.refactor.rename_module.ex). 
I added tests. The code is structured to drop into Igniter.Refactors.Rename with a module rename.

saschabrink:
Closes #373

A few notes to make review easier:

**Code style**
`mix format` has been run. Tests follow the same naming pattern as the existing `rename_function` suite (`lib/example.ex`, `lib/some_module.ex`, etc.). The new tests are grouped in a `describe "rename_module"` block — that was a deliberate choice to keep them visually separated, but happy to flatten them into the existing list if you prefer consistency.

**Why the longer `@doc` than `rename_function`**
I spent a while working through the edge cases and the doc reflects that. Everything static is handled correctly: plain `alias`, multi-alias (`alias Foo.{Bar, Other}`), `alias Foo.Bar, as: B`, submodules, test modules, string literals, and file moves. The longer doc makes the alias behaviour explicit upfront so callers know exactly what to expect without reading the source.

**Test modules are inferred and renamed together**
`FooTest` is inferred from the old module name and renamed alongside it as a convenience — no configuration needed. Happy to make this opt-out via an option if you'd prefer.

**Quick walkthrough through the code**
`rename_module` first breaks the module down into all the forms it needs — string parts, atom lists for the AST, short form, test module variant. The logic then flows as a pipe:

- `rewrite_affected_files` — walks all files that mention the module and runs two passes per file: first AST (`defmodule`, `alias`, `use`, call sites, short forms via `expand_alias`), then a string replace for literals and comments
- `move_submodule_files` — moves any files whose path sits under the old module directory (e.g. `lib/example/worker.ex` → `lib/new_example/worker.ex`)
- `move_module_file` — moves the module's own file

# Contributor checklist

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [X] Documentation changes
- [X] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/igniter/refactors/rename.ex
+++ b/lib/igniter/refactors/rename.ex
@@ -1049,8 +1049,7 @@ defmodule Igniter.Refactors.Rename do
              fn
                %Zipper{
                  node:
-                   {:alias, _,
-                    [{{:., _, [{:__aliases__, _, ^namespace_segs}, :{}]}, _, members}]}
+                   {:alias, _, [{{:., _, [{:__aliases__, _, ^namespace_segs}, :{}]}, _, members}]}
                } ->
                  Enum.any?(members, fn
                    {:__aliases__, _, ^old_short} -> true
@@ -1064,8 +1063,7 @@ defmodule Igniter.Refactors.Rename do
                   node:
                     {:alias, alias_meta,
                      [
-                       {{:., dot_meta, [{:__aliases__, ns_meta, _}, :{}]}, brace_meta,
-                        members}
+                       {{:., dot_meta, [{:__aliases__, ns_meta, _}, :{}]}, brace_meta, members}
                      ]}
                 } ->
                remaining =
--- a/lib/igniter/refactors/rename.ex
+++ b/lib/igniter/refactors/rename.ex
@@ -979,8 +979,10 @@ defmodule Igniter.Refactors.Rename do
       igniter
       |> replace_module_strings_in_file(path, old_str, new_str)
       |> Igniter.update_elixir_file(path, fn zipper ->
+        {_, new_aliases} = hd(renames)
+
         zipper
-        |> rename_aliased_short_forms(old_aliases, old_short, new_short)
+        |> rename_aliased_short_forms(old_aliases, new_aliases, old_short, new_short)
         |> apply_module_renames(renames)
         |> then(&{:ok, &1})
       end)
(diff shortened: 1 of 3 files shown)


lib/igniter/refactors/rename.ex
```diff
@@ -860,4 +932,196 @@ defmodule Igniter.Refactors.Rename do
     |> Module.split()
     |> Enum.map(&String.to_atom/1)
   end
+
+  ## rename_module helpers
+
+  defp find_module_file(igniter, module_str) do
```
zachdaniel: There are existing utilities for finding the file a module is in.

lib/igniter/refactors/rename.ex
```diff
@@ -860,4 +932,196 @@ defmodule Igniter.Refactors.Rename do
...
+    end)
+    |> Enum.map(&Rewrite.Source.get(&1, :path))
+  end
+
+  # Detects `alias Foo.{Bar, Baz}` style references where the full module name
+  # (e.g. "Foo.Bar") is not present as a substring. Checks for the namespace
+  # prefix followed by ".{" and the short segment name anywhere in the content.
+  # This is a heuristic and may include false positives, but the subsequent AST
+  # passes are no-ops when nothing actually matches.
+  defp multi_alias_in_content?(content, old_parts) do
```
zachdaniel: This should be using the AST and the zipper ideally not string matching.

zachdaniel: I guess based on the content this is meant as a performance optimization?

zachdaniel (commented):
Looking mostly
Good have a few comments!

saschabrink:
Thanks for the review! Pushed an update:

1. **`find_module_file`** — replaced with `Igniter.Project.Module.find_module/2`. Should've used the existing utility — sorry for the duplication.

 2. **`multi_alias_in_content?` (string matching)** — removed. The multi-alias detection now runs on the AST via a small `has_multi_alias_for?` helper that zippers for `{:alias, _, [{{:., _, [{:__aliases__, _, ns}, :{}]}, _, shorts}]}`, which is reused both in `find_affected_files` and in the short-form renamer.

3. **String pre-filter in `find_affected_files`** — yes, this is a performance filter, now explicitly: cheap `String.contains?` for the full module name (covers `defmodule`, dotted refs, string literals), with an AST fallback guarded by a cheap `"<Namespace>.{"` substring check so we don't parse every `.ex` file in the project for a multi-alias that almost never exists. False positives here are harmless — the downstream AST pass is a no-op when nothing matches.

---

What I already started before your comment:
- Fixed an ordering bug where running the AST pass before the string replace caused `Example` → `NewExample` to cascade into `NewNewExample` (the re-rendered content contained `NewExample`, which then matched `Example` again). Just had this in one of my projects. Rename module is not the easiest to implement...
- Added `import` and `require` tests for `rename_module`. Worked before but was untested. 

---

Tests:
........................
Finished in 0.2 seconds (0.00s async, 0.2s sync)
24 tests, 0 failures

saschabrink:
Fixed another bug in multi-alias handling: when renaming two LiveView pages from a shared namespace to separate ones — e.g. alias MyApp.{FooLive, BarLive} after FooLive → MyApp.Foo.IndexLive and BarLive → MyApp.Bar.IndexLive — the brace form would collapse to alias MyApp.{IndexLive, IndexLive} instead of expanding to two separate alias lines. 

Personally, I'm not a fan of the brace alias form — it's always been a source of friction. That said, I spent another hour specifically hunting for edge cases in the brace handling and found the one above, which is now covered. I hope this will save another developer some headache.

zachdaniel (approved):

zachdaniel:

saschabrink:
I have to thank you so much for this wonderful project and the super quick responses!

PR #377, opened 2026-05-09
fix: don't drop --key=value flags when value contains a dot

shahryarjb:
> **I may be wrong, please do not consider this PR, but in my tests on my personal project, this problem was solved with this PR.**
 
`args_for_group/2` strips other-group namespaced flags (`--other_group.foo`) by checking `String.contains?(arg, ".")`, but `arg` is the whole `key=value` string — so any `--key=value` whose value has a dot gets silently dropped before reaching OptionParser.

  Affects every `:float`, paths, module names, IPs, etc.

      $ mix foo --rate=1.5            # before: []   after: [rate: 1.5]
      $ mix foo --out=priv/foo.json   # before: []   after: [out: "priv/foo.json"]
      $ mix foo --rate 1.5            # already worked (two-arg form)

  Fix: split on `=` first, only check the flag-name portion for `.`. Adds 14
  unit tests for `args_for_group/2` (none existed) and 10 end-to-end tests
  covering `:string`, `:float`, `:integer`, `:boolean`, `:keep`, `:csv` —
  prior coverage was `:string` only.

# Contributor checklist

- [ x Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/igniter/util/info.ex
+++ b/lib/igniter/util/info.ex
@@ -389,7 +389,7 @@ defmodule Igniter.Util.Info do
           _ ->
             rest_args = args_for_group(["-" <> v2 | rest], group)
 
-            if String.contains?(arg, ".") do
+            if flag_namespaced?(arg) do
               rest_args
             else
               ["--#{arg}"] ++ rest_args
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #379, opened 2026-05-28
Enhancement: adding hex confirmation for igniter.install

### LINKED ISSUE #125: Confirm that the user is installing the correct package before adding it to the mix.exs file
**Is your feature request related to a problem? Please describe.**

My thought is around a possible attack vector and how we can mitigate it. If you make a typo eg. `mix igniter.install pheonix` and a bad actor is squatting that package, you could download a malicious package and have malicious code automatically run in an installer.

**Describe the solution you'd like**

One way I thought of helping with this is to show details of the package to be installed, even before the 'confirm adding this to mix.exs'. For hex packages, Igniter has to fetch details of the package to figure out what version to add to mix.exs, so it would also fetch details like description, authors, publish date, # of downloads of the version/all versions, etc.

We could have a nice summary screen for users to confirm that yes, this is the correct package they really want to install. eg.

```
You are installing the package "ash_phoenix":

Description:       Utilities for integrating Ash and Phoenix
Current version:   2.1.8 (released Oct 29, 2024)
hex.pm authors:    ash-project
hex.pm publishers: ash-project
Dependencies:      ash, phoenix, phoenix_html, phoenix_live_view
Downloads:         164 (this version), 3 850 (last 7 days), 229 751 (all time)

Is this correct? (y/N)
```

For non-hex packages, I'm not sure what you could do. Maybe fetch the mix.exs file from the project root (eg. on GitHub) and read it, but that could come with its own security issues?

ESmithByui:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [x] Update dependencies

# Summary
Closes #125 

Creates a confirmation section that matches the proposed solution in the above issue.
Only difference is instead of (y/N), it's (Y/n)

Dependencies changed to change decimal to 3.0. Jason and Doctor were the ones relying on decimal 2.0. When running `mix check` decimal 2.0 came back with vulnerabilities.


zachdaniel (changes requested):
Can we add a test for this? We could make it a test that doesn't run by default for example, or we could stub out the response from hex and just test the presentation behavior.

ESmithByui:
Sure, I'll get to work on this over the weekend.

ESmithByui:
> Can we add a test for this? We could make it a test that doesn't run by default for example, or we could stub out the response from hex and just test the presentation behavior.

Tests are now added in for it.

zachdaniel:
Mind looking into the failing checks here?

ESmithByui:
I'm reading over the errors, I think it's because of the dependency changes. From what I can tell whatever is checking is trying to use the old dependencies and then failing because of the updates to jason, decimal and doctor.

ESmithByui:
I think the cache is restoring an old version of either the mix.exs or the lock. Then it runs compile and the versions differ. I'm not sure how the CI works, but from what I see in multiple of these sailing checks is the dependencies don't match after the cache is restored.

zachdaniel:
Ah, right. I think it should be fixed for future builds. Thanks!

ESmithByui:
No problem. And because of this I learned more about the CI. Win win.

zachdaniel:

PR #387, opened 2026-06-28
fix: resolve Elixir 1.20 type-check warnings

gilbertwong96:
- `Igniter.Test.assert_has_issue/3` passed `%Igniter{}` to `Rewrite.source/2`, which expects `%Rewrite{}`; use `igniter.rewrite`. This also fixes a latent bug and resolves the cascading unknown-key warning on `igniter.issues`.
- `Igniter.Util.Info.add_deps/3` interpolated the dep tuple directly; use the bound dep name atom so the suggested install command is valid.

### DIFF
--- a/lib/igniter/test.ex
+++ b/lib/igniter/test.ex
@@ -424,7 +424,7 @@ defmodule Igniter.Test do
       end
 
     if path do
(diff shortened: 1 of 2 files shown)


zachdaniel:

PR #399, opened 2026-09-02
fix: Control rerunning installers for existing packages

### LINKED ISSUE #66: Support not running installers if a package is already present
**Describe the bug**
Right now, if a package is already installed, Igniter outputs that the dependency is already in `mix.exs` but proceeds to run the installer:

```
Dependency ash_postgres is already in mix.exs with the desired version. Skipping.

The following installer was found and executed: `ash_postgres.install`:

    No proposed content changes!

The following tasks will be run

* ash.codegen initialize
```

**To Reproduce**
Run `mix igniter.install ash_postgres` twice.

**Expected behavior**
To avoid inconsistency (i.e. saying "skipping" then proceeding to install), if a package is already installed, I'd suggest either:
- Skipping the installation as well by default, but supporting a flag like to run the installer anyway even if present.
- Avoid saying "Skipping" by default, always running the installer, but support `--skip-if-present` which would skip the installer if passed.

mvanhorn:
In `Igniter.Util.Install`, identify requested dependency names that are present in the incoming Igniter project before dependency composition changes the project, then use that set when selecting available installer tasks. For an existing dependency, run its installer when `--yes` (or the existing non-TTY automatic-yes behavior) applies; otherwise prompt interactively, while `--skip-installed` filters it out without prompting. Add `skip_installed` to the install-specific option parsing in both the normal Mix task and the copied/archive task path, and document the switch in the canonical task help.

`mix igniter.install` currently reports that an already-present dependency is being skipped, but it still discovers and runs that dependency's installer. This makes repeat invocations surprising and prevents callers such as Fireside from requesting "install only when newly added" behavior. The maintainer-directed behavior is to ask whether an existing package's installer should be rerun, treat `--yes` as acceptance, and provide `--skip-installed` to suppress both the prompt and rerun. The bundle shows no assignee, claim, competing PR, or closed prior attempt.

Installing a package that is absent from `mix.exs` runs its discovered installer without asking the rerun question.
Reinstalling an already-present package interactively and answering yes reruns its installer; answering no leaves the installer unexecuted.
Reinstalling an already-present package with `--yes` reruns its installer without waiting for input.
Reinstalling an already-present package with `--skip-installed` does not prompt and does not run its installer, including when `--yes` is also supplied so the explicit skip wins.
A mixed request containing new and existing packages skips only the existing package's installer under `--skip-installed` and still runs the new package's installer.

# Contributor checklist

- [ ] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring

Fixes #66


zachdaniel:

PR #400, opened 2026-09-03
fix: restore application env exactly after evaluating project config

neilberkman:
`with_evaled_configs/2` evaluates the project's `config/config.exs` and `config/<env>.exs`, applies the result with `Application.put_all_env/1`, and afterwards restores a per-application snapshot with `Application.put_all_env/1` again. `put_all_env` only writes the keys it is given, so any key the evaluated config introduced that was absent before stays set in the VM after the call returns.

This shows up in test suites that drive an Igniter installer against fixtures configuring the library under test. Every task that touches files goes through `Igniter.format/2`, so after such a test the fixture's `config :my_lib, MyLib.Config, ...` remains under `{:my_lib, MyLib.Config}` for the rest of the run, and unrelated tests that read that key pass or fail depending on test order.

The restore now deletes keys that were not in the snapshot before putting the snapshot back. The regression test covers a key that did not exist before, a pre-existing key that the evaluated config assigned a different value, and an application that had no environment at all.

`test/mix/tasks/igniter.install_test.exs` on current main does not compile (`install/1` is undefined); this change does not touch it.


zachdaniel:

